PR #3, opened 2020-04-03
Jon asked me to revert his implementation to fix omp_lock

estewart08:


estewart08:
Wanted to make sure Jon and others still wanted this in aomp 11.

JonChesterfield:
The clock/spinlock version is worse than the one based on sleep.

Trunk prints "unsupported", but that breaks critical sections which are otherwise OK.

No version actually provides a per-thread lock

I think we should move to the per-warp lock based on sleep for aomp-11.

ronlieb:
after I made the fix for asserter_ok, which modified the code generation for __kmpc_active)thread)mask and __kmpc_syncwarp to both emit a 64 bit type for parameter and return type, I noticed that omp_lock test started passing. 

in light of the above i am not sure we need to revert this patch.

The fix I made was related to the patch below that came in from upstream,
seems like its fixing a similar problem.

commit 2cd7fafc11fc1f721e88257d9befcc36ccf54294
Author: Alexey Bataev <a.bataev@hotmail.com>
Date:   Mon Aug 26 19:07:48 2019 +0000

    [OPENMP][NVPTX]Fix critical region codegen.
    
    Summary:
    Previously critical regions were emitted with the barrier making it a
    worksharing construct though it is not. Also, it leads to incorrect
    behavior in Cuda9+. Patch fixes this problem.

estewart08:
omp_lock should be passing because these patches have been applied by Jon already. He is saying that this solution should be done another way and we should revert these proposed already applied patches.

JonChesterfield:
I like the s_sleep version. It works for critical and as well as the clock version for locks

ronlieb:
On April 3, your test report stated that omp_lock is failing. So is  this schroedinger's test case? it passes and it fails ?

JonChesterfield:
I've somewhat lost the thread of patches and branches, but can say amd-stg-openmp has the implementation of amdgcn_locks.hip that I think we should be shipping.

estewart08:
Ok, closing the request.

PR #5, opened 2020-04-10
 only print ptxas messages for CMAKE_VERBOSE

gregrodgers:
Jon, I am creating a pr that will cleanup all the useless ptxas messages unless you turn on CMAKE_VERBOSE.   This is fairly simple so can you get this upstream?   Thanks.

### DIFF
--- a/openmp/libomptarget/deviceRTLs/nvptx/CMakeLists.txt
+++ b/openmp/libomptarget/deviceRTLs/nvptx/CMakeLists.txt
@@ -86,7 +86,11 @@ if(LIBOMPTARGET_DEP_CUDA_FOUND)
   set(LIBOMPTARGET_NVPTX_DEBUG FALSE CACHE BOOL
     "Activate NVPTX device RTL debug messages.")
   if(${LIBOMPTARGET_NVPTX_DEBUG})
-    set(CUDA_DEBUG -DOMPTARGET_NVPTX_DEBUG=-1 -g --ptxas-options=-v)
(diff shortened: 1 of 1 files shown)


JonChesterfield:
I'll take a look when I get back from leave. It looks like that option makes the debug nvptx build noisy, but that's presumably what the nvptx devs wanted.

gregrodgers:
This is not part of the debug build.  The release build gets these messages and if you have cuda installed, the default aomp build will build the nvptx deviceRTLs.    I am trying to cleanup the build logs to make them more meaningful.  This patch allows developers to turn them back on with verbose cmake.   Please squash and merge this.    Thank you.

JonChesterfield:
By debug build I mean with LIBOMPTARGET_NVPTX_DEBUG set. I think the verbose option is responsible for about 600 lines of noise in the build output. Raised https://reviews.llvm.org/D78334

That patch doesn't include the conditional branch on CMAKE_VERBOSE in the hope that it, or an equivalent, won't be deemed necessary. Grep suggests CMAKE_VERBOSE isn't used for anything in llvm yet and I'd like to avoid that discussion.

gregrodgers:
I am going to put this into another commit. I recommend that the phab review https://reviews.llvm.org/D78334  should be changed to print ptx when CMAKE_VERBOSE is on.

JonChesterfield:
Forgot about this. George wanted a cmake flag to control it. I don't know how to add such and haven't done the stack overflow digging to find out.

PR #6, opened 2020-04-10
use -std=c++14 for both nvcc and clang compiles of nvptx deviceRTL

gregrodgers:
This removes lots of warnings about templates .  We also use c++14 when compiling with clang for amdgcn so we should do the same with nvptx.


gregrodgers:
We can drip this pr.  c++14 added to compiler options upstream.

PR #7, opened 2020-04-10
fix problem where atomicAdd of 0 for long long creates atom.or.b64 wh…

gregrodgers:
John, this is a bit of a hack but it gets my k4000 (sm_30) working.  The #if defined .. changes nothing for amdgcn and any cuda_arch >= 35.  This results in the generation of a ptx atom.or.b64.   With cuda_arch<35, the source indirectly adds a zero.  This avoids the generation of atom.or.b64 and the fail I was getting for sm_30 compiles.  We need this or something else that does not fail soon in amd-stg-openmp soon.

### DIFF
--- a/openmp/libomptarget/deviceRTLs/common/state-queue.h
+++ b/openmp/libomptarget/deviceRTLs/common/state-queue.h
@@ -37,7 +37,7 @@ private:
   INLINE static uint32_t ID(uint32_t ticket);
   INLINE bool IsServing(uint32_t slot, uint32_t id);
   INLINE void PushElement(uint32_t slot, ElementType *element);
-  INLINE ElementType *PopElement(uint32_t slot);
+  INLINE ElementType *PopElement(uint32_t slot, uint64_t * zero);
   INLINE void DoneServing(uint32_t slot, uint32_t id);
 
 public:
--- a/openmp/libomptarget/deviceRTLs/common/state-queuei.h
+++ b/openmp/libomptarget/deviceRTLs/common/state-queuei.h
@@ -51,9 +51,16 @@ omptarget_nvptx_Queue<ElementType, SIZE>::PushElement(uint32_t slot,
 
 template <typename ElementType, uint32_t SIZE>
 INLINE ElementType *
-omptarget_nvptx_Queue<ElementType, SIZE>::PopElement(uint32_t slot) {
+omptarget_nvptx_Queue<ElementType, SIZE>::PopElement(uint32_t slot,
+  uint64_t * zero_ptr) {
+#if defined(__AMDGCN__) || ( defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 350 )
   return (ElementType *)__kmpc_atomic_add(
       (unsigned long long *)&elementQueue[slot], (unsigned long long)0);
+#else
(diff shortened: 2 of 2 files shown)


JonChesterfield:
This isn't very pretty. It only works because nvcc does such a poor job of inlining that it doesn't notice the indirection. Changing the interface to work around old hardware isn't ideal either.

Can we isolate it more aggressively? Introduce a function in the source marked noinline that does the pass in zero trickery, all hidden behind the ifdef?

I still think the better answer is to move you to a newer nvptx card. March 2013 was a long time ago, and I'm not confident this will be the last of the bugs we hit on it.

JonChesterfield:
For example, does implementing this as the following also work for you:
```
#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 350
static uint64_t __kmpc_atomic_add_zero(uint64_t *address)
{
  // atomic add of constant zero creates invalid atom.or.b64 for old cuda_archs with
  // some cuda toolchains, but those same toolchains also miss the interprocedural
  // optimisation which would undo this indirection
  uint64_t zero = 0;
  return __kmpc_atomic_add(address, &zero);
}
#endif 

#if defined(__AMDGCN__) || ( defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 350 )
  return (ElementType *)__kmpc_atomic_add(
      (unsigned long long *)&elementQueue[slot], (unsigned long long)0);
#else
  return (ElementType *)__kmpc_atomic_add_zero(
      (unsigned long long *)&elementQueue[slot]);
#endif
```

gregrodgers:
I did find another atomicAdd of a 64 bit zero in another test case. I wish I could figure out how to turn this off in the ptx generation.

JonChesterfield:
Doesn't make the patch much nicer but we could check for the zero at runtime and dispatch through an intermediate function. Or, probably better, just use inline asm for __kmpc_atomic_add. This can't go upstream anyway - the last patch that wanted to work around deficiencies in the cuda toolchain ultimately failed, and sm_30 is too slow to be of interest in production.

gregrodgers:
I am going to delete this pr.  I have a cleaner hack.

diff --git a/openmp/libomptarget/deviceRTLs/common/target_atomic.h b/openmp/libomptarget/deviceRTLs/common/target_atomic.h
index 3c905d3..9dc3211 100644
--- a/openmp/libomptarget/deviceRTLs/common/target_atomic.h
+++ b/openmp/libomptarget/deviceRTLs/common/target_atomic.h
@@ -16,7 +16,14 @@
 #include "target_impl.h"

 template <typename T> INLINE T __kmpc_atomic_add(T *address, T val) {
+#if defined(__AMDGCN__) || ( defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 320 ) || !defined(__clang__)
   return atomicAdd(address, val);
+#else
+  //  Hack for k4000 without a working  64bit atomic add
+  uint newval = (uint) val;
+  uint* uintaddr = (uint *) address; // take little end for 32 bit atomic op
+  return (T) atomicAdd(uintaddr, newval);
+#endif
 }

 template <typename T> INLINE T __kmpc_atomic_inc(T *address, T val) {

PR #9, opened 2020-04-11
Inform user if global symbol not found when loading image

ronlieb:


ronlieb:
this is motiviated by simple cases where a static global inside a declare target does not resolve.
about to push in the test case  test/smoke/targ_static

PR #11, opened 2020-04-16
Pr for jon sm 30

gregrodgers:
This is another hack but better than the last.   Lets discuss.


gregrodgers:
This was another mistake against the wrong branch

PR #13, opened 2020-04-18
Fixes for make check-all as encountered in testing for amd-stg-open

ronlieb:
All of these changes were tested against a clean build and full battery
of tests.   smoke, examples, spec, openmpapps, nekbone, sollve

### DIFF
--- a/clang/lib/Basic/TargetInfo.cpp
+++ b/clang/lib/Basic/TargetInfo.cpp
@@ -33,9 +33,9 @@ TargetInfo::TargetInfo(const llvm::Triple &T) : TargetOpts(), Triple(T) {
   TLSSupported = true;
   VLASupported = true;
   NoAsmVariants = false;
-  HasLegalHalfType = true;
+  HasLegalHalfType = false;
   HasFloat128 = false;
-  HasFloat16 = true;
+  HasFloat16 = false;
   PointerWidth = PointerAlign = 32;
   BoolWidth = BoolAlign = 8;
   IntWidth = IntAlign = 32;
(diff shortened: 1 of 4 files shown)


JonChesterfield:
I can't tell which branch this is against. amd-stg or amd-stg-openmp?

The changes to float16 and TLSSupported are worrying - are these things that are enabled but broken in the amd-stg branch? What's the corresponding upstream code?

Amdgcn macro renaming is uncontentious, though it's weird that the branch doesn't match trunk there.

ronlieb:
This is against amd-stg-openmp , its the set of changes that I had to make to the 1st patch that was pushed to gerrit to keep the lit failures from changing. It also includes the changes I needed to make to a 2nd patch that has not been pushed to gerrit yet.

all of these changes before fixing them, introduced lit failures when applied to amd-stg-open.  I needed to verify that the fixes needed in amd-stg-open would not break if applied later to amd-stg-openmp , which then meant I could just push them in since I had verified them via build/test, so I am doing so.

JonChesterfield:
I remain confused. Does amd-stg pass make check? I can't think of a reason why the openmp branch would want to claim TLS works, or float16 doesn't. VLA is borderline as we might be using them and depending on an optimisation to remove them before codegen

ronlieb:
with these same changes applied to patch 1 on gerrit, and patch 2 pending for gerrit. make check-all passes (3 lit failures from before our patches are still present)  

The TLA and FLOAT changes were stuff we injected into aomp about a year ago to fix problems at the time and are not needed any more (in my opinion). I think upstream fixed the issues differently.

JonChesterfield:
I guess we're going to have a long list of local patches that need to be painstakingly checked against upstream and the amd-stg branch.

So right now, amd-stg-openmp passes our tests, but we need these patches as part of trying to keep amd-stg also passing some tests while merging in some of our code?

Might need to discuss this on on Skype next week. I don't think I follow the goal here. Making changes that look wrong in order to balance out other local changes doesn't seem viable.

JonChesterfield:
Related - when trunk and amd-stg differ on code that interacts with openmp, which one do we intend to follow? That seems a different question to whether we periodically rebase on trunk or on amd-stg

PR #14, opened 2020-04-20
Let InitTempAlloca accept addrspace cast allocas

JonChesterfield:
remove if(amd) workarounds from codegen

InitTempAlloca does the right thing for allocas in non-default address spaces, it just has too aggressive a precondition. Intend to propose this as a patch against trunk shortly.


JonChesterfield:
Proposed D78495 to remove the patch against clang

PR #17, opened 2020-04-20
Slightly simplify diff against trunk

JonChesterfield:
I'm intending to write 'new CGOpenMPRuntimeAMDGCN' in place of 'new CGOpenMPRuntimeNVPTX' in the near future, and this patch will help me keep the diff against trunk clearer.

### DIFF
--- a/clang/lib/CodeGen/CodeGenModule.cpp
+++ b/clang/lib/CodeGen/CodeGenModule.cpp
@@ -210,9 +210,13 @@ void CodeGenModule::createOpenMPRuntime() {
   switch (getTriple().getArch()) {
   case llvm::Triple::nvptx:
   case llvm::Triple::nvptx64:
-  case llvm::Triple::amdgcn:
     assert(getLangOpts().OpenMPIsDevice &&
            "OpenMP NVPTX is only prepared to deal with device code.");
+    OpenMPRuntime.reset(new CGOpenMPRuntimeNVPTX(*this));
+    break;
+  case llvm::Triple::amdgcn:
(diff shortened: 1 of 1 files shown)


clang/lib/CodeGen/CodeGenModule.cpp
```diff
@@ -210,6 +210,10 @@ void CodeGenModule::createOpenMPRuntime() {
   switch (getTriple().getArch()) {
   case llvm::Triple::nvptx:
   case llvm::Triple::nvptx64:
+    assert(getLangOpts().OpenMPIsDevice &&
+           "OpenMP NVPTX is only prepared to deal with device code.");
+    OpenMPRuntime.reset(new CGOpenMPRuntimeNVPTX(*this));
+    break;
   case llvm::Triple::amdgcn:
     assert(getLangOpts().OpenMPIsDevice &&
            "OpenMP NVPTX is only prepared to deal with device code.");
```
ronlieb, line 219: should this line say AMDGCN instead of NVPTX ?

JonChesterfield: I was thinking of changing both at the same time, once said class exists. Doesn't make much odds though - I'll change this diff up front

ronlieb:
Lgtm

clang/lib/CodeGen/CodeGenModule.cpp
JonChesterfield: Done, though github is only showing the change if I click through to the commit

PR #18, opened 2020-04-20
Remove hip_host_runtime_api.h

ronlieb:
Removes 3100+ lines of difference relative to trunk llvm
builds, passes lotso tests


ronlieb:
This patch DOES remove the file completely.

look closely at:

 3,033 clang/lib/Headers/hip/hip_host_runtime_api.h

PR #20, opened 2020-04-21
Remove dead files

JonChesterfield:
These are not present in head of amd-stg-open, and llvm builds and passes our tests with them removed. Transforms removed were under HC.


JonChesterfield:
They are not present in llvm trunk, I haven't checked whether they were transiently upstream.

PR #27, opened 2020-04-23
Initialize IsNew and Pointer_IsNew in target_data_begin

ronlieb:
These two locals were uninitialized at delcaration, and then passed by
reference to Device.getOrAllocTgtPtr which in turn did not assign on all
paths within the fucntion.

This caused Nekbone to fail at runtime along
with its reduced test case map_zero_bug.

Needs to be upstreamed.


JonChesterfield:
Nice find

PR #28, opened 2020-04-23
This pr removes atmi and bring in necessary atmi functionality into t…

gregrodgers:
…he hsa plugin implementation.  It also merges in hostcall functionality into the impl and adds the device runtime library build into deviceRTLs/hostcall

Associated with this pr is a change to aomp scripts build_aomp,sh and build_openmp.sh
Also, you must not build hostcall in aomp-extras.  So the third pr is need for aomp-extras.  We should continue to try to remove functionality from aomp-extras.


JonChesterfield:
That's a really big change. 12.5k! That can be cut down dramatically by removing more of the atmi infra.

Is this copying files out of atmi/extras, or is it also changing them in flight? OK with me if it's moving files unchanged, can't really review it if there are also functional changes mixed in.

llvm/lib/Support/CommandLine.cpp
```diff
@@ -2466,7 +2466,7 @@ class VersionPrinter {
 #ifdef PACKAGE_VENDOR
     OS << PACKAGE_VENDOR << " ";
 #else
-    OS << "LLVM (http://llvm.org/):\n  ";
+    OS << "AOMP-11.5-0 (http://github.com/ROCm-Developer-Tools/aomp):\n Source ID:11.5-0-871be7a4290657aa2ad19a1f3198b91952e29c00\n  ";
```
JonChesterfield, line 2469: This is usually left out of commits

openmp/libomptarget/deviceRTLs/CMakeLists.txt
```diff
@@ -12,3 +12,4 @@
 
 add_subdirectory(amdgcn)
 add_subdirectory(nvptx)
+add_subdirectory(hostcall)
```
JonChesterfield, line 15: I don't think hostcall works with nvptx, so we'll probably need to move it under amdgcn before upstreaming.

openmp/libomptarget/plugins/hsa/impl/amd_hostcall.h
```diff
@@ -0,0 +1,281 @@
...
+#define AMD_HOSTCALL_H
+
+#include <stddef.h>
+#include <stdint.h>
+
+#ifndef AMD_HOSTCALL_EXPORT_DECORATOR
+#ifdef __GNUC__
+#define AMD_HOSTCALL_EXPORT_DECORATOR __attribute__((visibility("default")))
+#else
+#define AMD_HOSTCALL_EXPORT_DECORATOR __declspec(dllexport)
```
JonChesterfield, line 11: This sort of stuff can go post commit. Marking every function with a macro is crude relative to using an export list when building the library, and thankfully we don't need to build on windows.

llvm/lib/Support/CommandLine.cpp
gregrodgers: ARG!  Sorry about the automated CommandLine change getting included.   How can I remove this from the pull request patch? 

Yeah, this was most of atmi runtime brought in without changing any of the files.   I agree, we can start stripping out pieces we don't need.    But remember, as I said in my email, Ashwin is going to help us remove dependency on comgr.     It was also most of the hostcall stuff merged without change from aomp-extras.

openmp/libomptarget/deviceRTLs/CMakeLists.txt
gregrodgers: umm, I wouldn't mind if we figured out a way for this to work.  Dont need it for printf but if we get serious about other glibc and/or mpi services, it would be good to have portable code.

JonChesterfield: I'd like that too. Not sure what the cuda apis for doing this look like, but I imagine it can be abstracted over.

llvm/lib/Support/CommandLine.cpp
JonChesterfield: I'd drop it from the patch locally then, `git commit --amend --no-edit && git push -f`. Github will adapt to the force-pushed branch.

ronlieb:
the time has come, this patch needs serious code reduction in order to ever get it upstreamed, or garroted.

JonChesterfield:
It can't be discarded. There's no alternative runtime library available. Going back to raw atmi, or to vdi, is not a step forward.

I've been sidetracked by replacing hostcall. That's somewhere between prototype and product - pretty sure the state machine is sound, API is almost good to go, want to do a bunch of stress testing and benchmarking.

Queue & kernel management should do done from rtl.cpp, instead of routing through the async plumbing. Suggest doing that starting at the kernel launch and working backwards.

Symbol management was badly delayed by the hsa vs comgr metadata mess but is no longer blocking.

Probably a few days work to cut out the async task dispatch and wire up the symbol tables more directly. Are you available to do so? I'm reluctant to drop hostcall before the concurrency model is totally pinned down as I'll never get the context back.

ronlieb:
i was not suggesting discarding this patch, just cleaning it up.

lets move the discussion into the teams chat , I started a note there about it. good topic for this mornings meeting.

PR #29, opened 2020-04-23
Duplicate ToolChains/HIP to allow divergence in implementation of OpenMP

JonChesterfield:
The new AMDGPUOpenMP toolchain is a copy of HIP with s/hip/amdgpuopenmp/g. The other changes are to use an instance of this for openmp, and to drop the HIP fat binary function which is unused.

This can then be followed by patches to the OpenMP toolchain to remove HIP specific things, and by patches to the HIP toolchain to remove OpenMP specific things, with the goal of making the HIP implementation coincide with that in amd-stg-open.

Some exact duplication will remain after that specialisation, which could be moved into helper functions in AMDGPU or similar, depending somewhat on how the corresponding reviews go.

To be totally explicit - this patch duplicates approx 750 lines of code in order to decouple the HIP and OpenMP implementations, reducing the risk of either breaking the other and hopefully facilitating moving openmp code into trunk or amd-stg-open.

### DIFF
--- a/clang/lib/Driver/CMakeLists.txt
+++ b/clang/lib/Driver/CMakeLists.txt
@@ -36,6 +36,7 @@ add_clang_library(clangDriver
   ToolChains/AIX.cpp
   ToolChains/Ananas.cpp
   ToolChains/AMDGPU.cpp
+  ToolChains/AMDGPUOpenMP.cpp
   ToolChains/AVR.cpp
   ToolChains/BareMetal.cpp
   ToolChains/Clang.cpp
--- a/clang/lib/Driver/Driver.cpp
+++ b/clang/lib/Driver/Driver.cpp
@@ -10,6 +10,7 @@
 #include "InputInfo.h"
 #include "ToolChains/AIX.h"
 #include "ToolChains/AMDGPU.h"
+#include "ToolChains/AMDGPUOpenMP.h"
 #include "ToolChains/AVR.h"
 #include "ToolChains/Ananas.h"
 #include "ToolChains/BareMetal.h"
@@ -754,7 +755,7 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
               auto &HIPTC =
                   ToolChains[HIPTriple.str() + "/" + HostTriple.str()];
               if (!HIPTC) {
-                HIPTC = std::make_unique<toolchains::HIPToolChain>(
+                HIPTC = std::make_unique<toolchains::AMDGPUOpenMPToolChain>(
                     *this, HIPTriple, *HostTC, C.getInputArgs(),
                     Action::OFK_OpenMP);
               }
--- a/clang/lib/Driver/ToolChains/AMDGPUOpenMP.cpp
+++ b/clang/lib/Driver/ToolChains/AMDGPUOpenMP.cpp
@@ -1,0 +1,604 @@
+//===- AMDGPUOpenMP.cpp - AMDGPUOpenMP ToolChain Implementation -*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "AMDGPUOpenMP.h"
+#include "AMDGPU.h"
+#include "CommonArgs.h"
+#include "InputInfo.h"
+#include "clang/Basic/Cuda.h"
+#include "clang/Driver/Compilation.h"
+#include "clang/Driver/Driver.h"
+#include "clang/Driver/DriverDiagnostic.h"
+#include "clang/Driver/Options.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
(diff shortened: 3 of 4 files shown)


JonChesterfield:
Passes other than snap, but snap is broken locally for me anyway. It's moving existing code around so most (maybe even all?) errors are found at compile time

ronlieb:
how is snap failing for you?

JonChesterfield:
Hangs. Haven't attempted to debug

ronlieb:
share the snap executable and I can try it on my gfx906 Box ?

ronlieb:
Hangs for me too…

#0  0x00007fffec1458fd in core::InterruptSignal::WaitRelaxed(hsa_signal_condition_t, long, unsigned long, hsa_wait_state_t) ()
   from /home/rlieberm/rocm/aomp_11.5-0/lib/libhsa-runtime64.so.1
#1  0x00007fffec1370f9 in HSA::hsa_signal_wait_relaxed(hsa_signal_s, hsa_signal_condition_t, long, unsigned long, hsa_wait_state_t) ()
   from /home/rlieberm/rocm/aomp_11.5-0/lib/libhsa-runtime64.so.1
#2  0x00007fffec129aea in amd::AqlQueue::~AqlQueue() ()
   from /home/rlieberm/rocm/aomp_11.5-0/lib/libhsa-runtime64.so.1
#3  0x00007fffec129d85 in amd::AqlQueue::Destroy() ()
   from /home/rlieberm/rocm/aomp_11.5-0/lib/libhsa-runtime64.so.1
#4  0x00007fffec135bc2 in HSA::hsa_queue_destroy(hsa_queue_s*) ()
   from /home/rlieberm/rocm/aomp_11.5-0/lib/libhsa-runtime64.so.1
#5  0x00007ffff4882480 in ATLProcessor::destroyQueues (this=<optimized out>)
    at /home/rlieberm/git/aomp11/amd-llvm-project/openmp/libomptarget/plugins/hsa/impl/machine.cpp:252
#6  0x00007ffff488a695 in core::Runtime::Finalize (this=<optimized out>)
    at /home/rlieberm/git/aomp11/amd-llvm-project/openmp/libomptarget/plugins/hsa/impl/system.cpp:334
#7  0x00007ffff48a1c2b in RTLDeviceInfoTy::~RTLDeviceInfoTy (
    this=0x7ffff4aafe00 <DeviceInfo>, __in_chrg=<optimized out>)
    at /home/rlieberm/git/aomp11/amd-llvm-project/openmp/libomptarget/plugins/hsa/src/rtl.cpp:358
#8  0x00007ffff66dc041 in __run_exit_handlers (status=0,

JonChesterfield:
Snap hangs without this patch as well, but there's no rush on landing it.

ronlieb:
which mpi did you build this with ? I think ethan had noticed that older MPIs would hang.
I will apply your patch and try it myself.

JonChesterfield:
Fairly elderly. It's what 18.04 comes with by default
```
ompi_info
                 Package: Open MPI buildd@lcy01-amd64-009 Distribution
                Open MPI: 2.1.1
  Open MPI repo revision: v2.1.0-100-ga2fdb5b
   Open MPI release date: May 10, 2017
```

This patch won't fix or cause a hang - it's a copy & rename of part of clang, and the hang looks like a race condition in one of the runtimes.

JonChesterfield:
I'm still sure we need something like this to progress clang codegen

ronlieb:

Probably, it was closed as a side effect of my idiotic repo fix.

Please reopen #5 and #37 as well?

ronlieb:
There are some upstream commits which affect the AMDGP and HIP Toolchains, These change where files are expected to be found. Lets circle back to this PR after the  commits are merged int amd-stg-openmp.  I expect those upstream changes to become stable in the next week or so, and then land in our branch around mid to late jun.

JonChesterfield:
Do you mean the archive patches? Hip were doing something unsound with static archives that I definitely don't want in openmp.

I still have no alternative ideas for progressing codegen for openmp in trunk, and don't have the go ahead from Greg. Blocked.

ronlieb:
i am referring to the patches that determin where to find libdevice bitcode files, what the names of these files are, and enforcing finding a ROCm install

JonChesterfield:
Ah, yes. That's also changing at present.

I'd actually rather split the toolchain before that lands as I suspect we'll want to do something slightly different, i.e. we probably don't want to require a whole rocm install. So keeping it in one file means control flow is going to get even worse.

ronlieb:
Since we will be part of the unified compiler as well as standalone, we will need to support both approaches of ROCm vs AOMP delivery.

JonChesterfield:
That's still different to the control flow in trunk and different to that in amd-stg-open, and the toolchain file isn't that simple to begin with.

ronlieb:
I grabbed this patch and applied it to our 11.6-0 build this morning, it applied clean, built clean, and has passed all our  tests   (smoke, omp5, hip-examples, examples, openmpapps, spec accel, spec hpc2020, sollve, nekbone). 

I also applied a variant of this to the 'patch #6'  for merge into amd-stg-open and it is behaving to my expectations.

In short, this patch works well, thanks, lets commit it.

JonChesterfield:
Thanks! I'm a little surprised this hadn't rotted.

PR #31, opened 2020-04-24
Extract code object version discovery into separate function

JonChesterfield:
and simplify error reporting

Not quite a non functional change in that it reduces the variety of failure messages. The new get_code_object_version could return a variety of error codes if necessary.

This patch would facilitate assuming v3 in release mode (checking in debug, or at build time, or not at all), but seems a reasonable change in it's own right.


JonChesterfield:
Accepted in the atmi branch, going to land it on llvm-project in the meantime

PR #33, opened 2020-04-24
support for a hostrpc service to call a host function with a function…

gregrodgers:
… pointer with no arguments.  hostrpc_fptr0() .  The demo of this is in examples/openmp/host_function.

### DIFF
--- a/openmp/libomptarget/deviceRTLs/hostcall/src/hostcall_service_id.h
+++ b/openmp/libomptarget/deviceRTLs/hostcall/src/hostcall_service_id.h
@@ -54,6 +54,7 @@ enum hostcall_service_id {
   HOSTCALL_SERVICE_MALLOC,
   HOSTCALL_SERVICE_FREE,
   HOSTCALL_SERVICE_DEMO,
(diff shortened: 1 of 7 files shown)


JonChesterfield:
I think I'd expect this to call the host function once per active warp. Is that the case?

ronlieb:

Hmm, the test case looks like it does it once per thread ?

gregrodgers:
It acts as if it is once per thread.  However, the internals do an aggregate rpc call once per warp.   To get it to appear to run once per warp you would have to put a mask on with an if statement.   On Friday, Ron and I talked about getting hostrpc_fptr0(fptr) to return an int but then that would require the function that the function pointer represents to always return an int so we are leaving this as is.  I am working on more powerful hostrpc_varfn(...) which would take a host function pointer but with a variable number of args.

gregrodgers:
Ron, please approve and merge this so we can continue cleanup effort on hostrpc and hsa/impl.

ronlieb:

I approved it 3 days ago, merged in now.

PR #34, opened 2020-04-24
Allow any archive file to be part of device compilation. Fixes isse #89.

jsjodin:
Archive files had to start with "lib" to be unbundled correctly. This patch removes this restriction.


JonChesterfield:
Thanks!

PR #36, opened 2020-04-28
Provide an inline version of comgr

JonChesterfield:
This is a proposed solution to ROCm wanting to use comgr everywhere while trunk wants a minimal, self contained plugin. It assumes AOMP also wants the minimal, self contained plugin. The same code could be used by ATMI.

The code is presently derived from comgr, aggressively cut down to the symbols used by the ATMI implementation that is currently used by the plugin. It uses exactly the same interface, so that compiling with a macro set will use amd_comgr.h instead. If this direction is acceptable, we can then replace llvm's elf library with libelf, binary support with a cut down metadata reader etc.

Putting the patch up now to check that this direction is considered reasonable. This makes comgr optional, but doesn't yet make the llvm libraries optional.

### DIFF
--- a/openmp/libomptarget/plugins/hsa/CMakeLists.txt
+++ b/openmp/libomptarget/plugins/hsa/CMakeLists.txt
@@ -76,6 +76,8 @@ include_directories(
 add_library(omptarget.rtl.hsa SHARED
       impl/atmi.cpp
       impl/atmi_interop_hsa.cpp
+      impl/comgr-metadata.cpp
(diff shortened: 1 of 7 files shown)


JonChesterfield:
Comgr is under the `Uni of Illinois` and libomptarget is under apache, which may mean I can't copy code from comgr into libomptarget even though AMD wrote both pieces.

JonChesterfield:
Abandoning. The license uncertainty is a step too far, will find an alternative path.

PR #37, opened 2020-04-29
use the clang generated symbol for each kernel ending in _kern_desc t…

gregrodgers:
…o store number of kernel args.  Then use this when initializing the runtime to calculate kernel_segment_size to avoid the need for comgr.  Remove the commented out printf in system.cpp to show comgr kss vs kss calculated from clang generated symbol.  It is up to Jonathan if he wants to use this instead of messagepack.   However, this PR shows how to pull any information from the kern_desc structure after code object is frozen.   This PR also moves the definition of the structure into an impl header with a note to keep in sync with clang. 

This PR does not get rid of comgr yet.  To do that two things are necessary.  1 remove this call at line 1606.    
 err = get_code_object_custom_metadata(types[i], tmp_module, module_size, gpu);

2. replace with equivalent function using message pack OR have the KernelInfoTable entries initialized in the symbol iterator at line 1465.

### DIFF
--- a/clang/lib/CodeGen/CGOpenMPRuntime.cpp
+++ b/clang/lib/CodeGen/CGOpenMPRuntime.cpp
@@ -4361,13 +4361,15 @@ void CGOpenMPRuntime::emitKmpRoutineEntryT(QualType KmpInt32Ty) {
 void CGOpenMPRuntime::emitStructureKernelDesc(CodeGenModule &CGM,
                                               StringRef Name, int16_t WG_Size,
                                               int8_t Mode, int8_t HostServices,
-                                              int8_t MaxParallelLevel) {
+                                              int8_t MaxParallelLevel,
+                                              int16_t num_args) {
 
   // Create all device images
   llvm::Constant *AttrData[] = {
-      llvm::ConstantInt::get(CGM.Int16Ty, 2), // Version
-      llvm::ConstantInt::get(CGM.Int16Ty, 9), // Size in bytes
+      llvm::ConstantInt::get(CGM.Int16Ty, 3),  // Version
+      llvm::ConstantInt::get(CGM.Int16Ty, 12), // Size in bytes
       llvm::ConstantInt::get(CGM.Int16Ty, WG_Size),
+      llvm::ConstantInt::get(CGM.Int16Ty, num_args),
       llvm::ConstantInt::get(CGM.Int8Ty, Mode), // 0 => SPMD, 1 => GENERIC
       llvm::ConstantInt::get(CGM.Int8Ty, HostServices), // 1 => use HostServices
       llvm::ConstantInt::get(CGM.Int8Ty, MaxParallelLevel)}; // number of nests
@@ -4389,6 +4391,7 @@ QualType CGOpenMPRuntime::getTgtAttributeStructQTy() {
     addFieldToRecordDecl(C, RD, KmpInt16Ty); // Version
     addFieldToRecordDecl(C, RD, KmpInt16Ty); // Struct Size in bytes.
     addFieldToRecordDecl(C, RD, KmpInt16Ty); // WG_size
+    addFieldToRecordDecl(C, RD, KmpInt16Ty); // num_args
     addFieldToRecordDecl(C, RD, KmpInt8Ty);  // Mode
     addFieldToRecordDecl(C, RD, KmpInt8Ty);  // HostServices
     addFieldToRecordDecl(C, RD, KmpInt8Ty);  // MaxParallelLevel
--- a/clang/lib/CodeGen/CGOpenMPRuntime.cpp
+++ b/clang/lib/CodeGen/CGOpenMPRuntime.cpp
@@ -4360,19 +4360,22 @@ void CGOpenMPRuntime::emitKmpRoutineEntryT(QualType KmpInt32Ty) {
 /// Emit structure descriptor for a kernel
 void CGOpenMPRuntime::emitStructureKernelDesc(CodeGenModule &CGM,
                                               StringRef Name, int16_t WG_Size,
-                                              int8_t Mode, int8_t HostServices,
-                                              int8_t MaxParallelLevel,
-                                              int16_t num_args) {
+                                              int16_t NumArgs, int8_t Mode,
+                                              int8_t HostServices,
+                                              int8_t MaxParallelLevel) {
 
   // Create all device images
+  int8_t byte_reserve = 0;
   llvm::Constant *AttrData[] = {
       llvm::ConstantInt::get(CGM.Int16Ty, 3),  // Version
       llvm::ConstantInt::get(CGM.Int16Ty, 12), // Size in bytes
       llvm::ConstantInt::get(CGM.Int16Ty, WG_Size),
-      llvm::ConstantInt::get(CGM.Int16Ty, num_args),
+      llvm::ConstantInt::get(CGM.Int16Ty, NumArgs),
       llvm::ConstantInt::get(CGM.Int8Ty, Mode), // 0 => SPMD, 1 => GENERIC
       llvm::ConstantInt::get(CGM.Int8Ty, HostServices), // 1 => use HostServices
-      llvm::ConstantInt::get(CGM.Int8Ty, MaxParallelLevel)}; // number of nests
+      llvm::ConstantInt::get(CGM.Int8Ty, MaxParallelLevel), // number of nests
+      llvm::ConstantInt::get(CGM.Int8Ty,
+                             byte_reserve)}; // reserved for fuuture use
 
   llvm::GlobalVariable *AttrImages = createGlobalStruct(
       CGM, getTgtAttributeStructQTy(), isDefaultLocationConstant(), AttrData,
@@ -4391,10 +4394,11 @@ QualType CGOpenMPRuntime::getTgtAttributeStructQTy() {
     addFieldToRecordDecl(C, RD, KmpInt16Ty); // Version
     addFieldToRecordDecl(C, RD, KmpInt16Ty); // Struct Size in bytes.
     addFieldToRecordDecl(C, RD, KmpInt16Ty); // WG_size
-    addFieldToRecordDecl(C, RD, KmpInt16Ty); // num_args
+    addFieldToRecordDecl(C, RD, KmpInt16Ty); // NumArgs
     addFieldToRecordDecl(C, RD, KmpInt8Ty);  // Mode
     addFieldToRecordDecl(C, RD, KmpInt8Ty);  // HostServices
     addFieldToRecordDecl(C, RD, KmpInt8Ty);  // MaxParallelLevel
+    addFieldToRecordDecl(C, RD, KmpInt8Ty);  // reserved
     RD->completeDefinition();
     TgtAttributeStructQTy = C.getRecordType(RD);
   }
--- a/clang/lib/CodeGen/CGOpenMPRuntime.h
+++ b/clang/lib/CodeGen/CGOpenMPRuntime.h
@@ -1695,9 +1695,8 @@ public:
 
   /// Emit structure descriptor for a kernel
   void emitStructureKernelDesc(CodeGenModule &CGM, StringRef Name,
-                               int16_t WG_Size, int8_t Mode,
-                               int8_t HostServices, int8_t MaxParallelLevel,
-                               int16_t num_args);
+                               int16_t WG_Size, int16_t NumArgs, int8_t Mode,
+                               int8_t HostServices, int8_t MaxParallelLevel);
 
   /// Emits OpenMP-specific function prolog.
   /// Required for device constructs.
--- a/clang/lib/CodeGen/CGOpenMPRuntime.h
+++ b/clang/lib/CodeGen/CGOpenMPRuntime.h
@@ -1696,7 +1696,8 @@ public:
   /// Emit structure descriptor for a kernel
   void emitStructureKernelDesc(CodeGenModule &CGM, StringRef Name,
                                int16_t WG_Size, int8_t Mode,
-                               int8_t HostServices, int8_t MaxParallelLevel);
+                               int8_t HostServices, int8_t MaxParallelLevel,
+                               int16_t num_args);
 
   /// Emits OpenMP-specific function prolog.
   /// Required for device constructs.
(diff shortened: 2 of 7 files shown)


clang/lib/CodeGen/CGOpenMPRuntime.cpp
```diff
@@ -4361,13 +4361,15 @@ void CGOpenMPRuntime::emitKmpRoutineEntryT(QualType KmpInt32Ty) {
...
-                                              int8_t MaxParallelLevel) {
+                                              int8_t MaxParallelLevel,
+                                              int16_t num_args) {
 
   // Create all device images
   llvm::Constant *AttrData[] = {
-      llvm::ConstantInt::get(CGM.Int16Ty, 2), // Version
-      llvm::ConstantInt::get(CGM.Int16Ty, 9), // Size in bytes
+      llvm::ConstantInt::get(CGM.Int16Ty, 3),  // Version
+      llvm::ConstantInt::get(CGM.Int16Ty, 12), // Size in bytes
```
ronlieb, line 4370: why isn't this 11 ?   have you run programs with this patch?

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
```diff
@@ -1329,10 +1329,10 @@ void CGOpenMPRuntimeNVPTX::GenerateMetaData(CodeGenModule &CGM,
...
   // Emit a kernel descriptor for runtime.
   StringRef KernDescName = OutlinedFn->getName();
-  CGOpenMPRuntime::emitStructureKernelDesc(CGM, KernDescName, FlatAttr,
-                                           IsGeneric,
-                                           1, // Uses HostServices
-                                           MaxParallelLevel);
+  CGOpenMPRuntime::emitStructureKernelDesc(
+      CGM, KernDescName, FlatAttr, IsGeneric,
+      1, // Uses HostServices
+      MaxParallelLevel, (int16_t)OutlinedFn->arg_size());
```
ronlieb, line 1335: This arg is added after the hos services arg, but the structure definition defines it to be two fields earlier

clang/lib/CodeGen/CGOpenMPRuntime.cpp
gregrodgers: Yes, the size of the struct is 12.  There is a typedef atmi_kern_desc_t for the struct.   Yes, this code works and if you uncomment those two printfs in system.cpp you will see that kernel_segment_size gets the same value calculated after comgr discovers the args and then adds in the implicit args size.

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
gregrodgers: That is an arg to the clang function emitStructureKernelDesc.  The actual generated structure order is inside emitStructureKernelDesc.

ronlieb:
one reason to retain the version field, is so you can emit a warning or error if the new runtimes are presented a kernel with an older version of this structure. The implication is that you cant reliably get the number_of_kernel args, I think because its not there in the KernDesc.

JonChesterfield:
I don't understand how this can be correct. The kernarg segment size is used to allocate space for the kernel arguments, and llc uses implicit arguments to implement some intrinsics, but the openmp layer can't know which implicit arguments will be introduced.

gregrodgers:
The runtime must fill in the implicit args.   ATMI does this for hostcall_ptr.   The rocr runtime will fill in other fields.  Here is the struct atmi uses to fill in the hostcall_ptr;
.typedef struct atmi_implicit_args_s {
  unsigned long offset_x;
  unsigned long offset_y;
  unsigned long offset_z;
  unsigned long hostcall_ptr;
  char num_gpu_queues;
  unsigned long gpu_queue_ptr;
  char num_cpu_queues;
  unsigned long cpu_worker_signals;
  unsigned long cpu_queue_ptr;
  unsigned long kernarg_template_ptr;
} atmi_implicit_args_t;

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
ashwinma: Maybe to avoid confusion, the args can be added in the same order to the `emitStructureKernelDesc` function as the structure?

clang/lib/CodeGen/CGOpenMPRuntime.h
```diff
@@ -1696,7 +1696,8 @@ class CGOpenMPRuntime {
   /// Emit structure descriptor for a kernel
   void emitStructureKernelDesc(CodeGenModule &CGM, StringRef Name,
                                int16_t WG_Size, int8_t Mode,
-                               int8_t HostServices, int8_t MaxParallelLevel);
+                               int8_t HostServices, int8_t MaxParallelLevel,
+                               int16_t num_args);
```
ashwinma, line 1700: (Minor) By going with the coding convention, this should be NumArgs I suppose

clang/lib/CodeGen/CGOpenMPRuntime.cpp
ashwinma: Why did it increase from 9 to 12 after you add 2 bytes? Is it some alignment/padding that is happening implicitly? If yes, we should add a dummy `reserved` field to make it more explicit.

clang/lib/CodeGen/CGOpenMPRuntime.cpp
```diff
@@ -4361,13 +4361,15 @@ void CGOpenMPRuntime::emitKmpRoutineEntryT(QualType KmpInt32Ty) {
 void CGOpenMPRuntime::emitStructureKernelDesc(CodeGenModule &CGM,
                                               StringRef Name, int16_t WG_Size,
                                               int8_t Mode, int8_t HostServices,
-                                              int8_t MaxParallelLevel) {
+                                              int8_t MaxParallelLevel,
+                                              int16_t num_args) {
```
ashwinma, line 4365: (Minor) By going with the coding convention, this should be NumArgs I suppose

openmp/libomptarget/plugins/hsa/impl/atmi_runtime.h
```diff
@@ -243,6 +243,22 @@ typedef struct atmi_kernel_s {
   uint64_t handle;
 } atmi_kernel_t;
 
+/**
```
ashwinma, line 246: I recommend moving this struct to atmi.h instead. 

atmi_runtime.h includes atmi.h anyway.

ashwinma:
> llc uses implicit arguments to implement some intrinsics

LLC needs offset_x, offset_y and offset_z to implement the intrinsics, which are also part of atmi_implicit_args_t and zeroed out by ATMI before kernel launch. If the LLC's requirements change, we will have to make those changes accordingly to atmi_implicit_args_t.

JonChesterfield:
Thanks for the info about implicit args - I had been wondering where the runtime sets it. Am I right in thinking that ATMI has chosen the position of fields in atmi_implicit_args_t to coincide with where llc currently expects to find the appropriate values?

I haven't looked through this patch in detail yet. I do think we want a convenient path to pass openmp specific information from the compiler to the runtime, probably per-kernel, and I quite like using a flat struct for that. Until we can specialise instances of the runtime machinery on a given kernel anyway.

openmp/libomptarget/plugins/hsa/impl/system.cpp
```diff
@@ -1516,6 +1518,41 @@ hsa_status_t populate_InfoTables(hsa_executable_t executable,
     register_allocation(reinterpret_cast<void *>(info.addr), (size_t)info.size,
                         place);
     SymbolInfoTable[gpu][std::string(name)] = info;
+
+    // Check if symbol is the kernel descriptor symbol, ends in "_kern_desc"
+    std::string sym_name(name);
+    std::size_t found = sym_name.rfind("_kern_desc");
```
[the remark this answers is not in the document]
gregrodgers: yes

clang/lib/CodeGen/CGOpenMPRuntime.h
gregrodgers: agree

clang/lib/CodeGen/CGOpenMPRuntime.cpp
gregrodgers: yes, added reserve field in last commit.  Size is still 12

openmp/libomptarget/plugins/hsa/impl/atmi_runtime.h
gregrodgers: This will all get moved around when we restructure into __himpl_

gregrodgers:
I appreciate all the feedback.   There is a lot of cleanup to do in plugins/hsa/impl , so lets merge this now.  I added a small change to this pr to move hostcall initialization into atmi init.  Now there are only atmi calls in rtl.cpp.

openmp/libomptarget/plugins/hsa/impl/atmi_runtime.h
gregrodgers: The uint fields are not  defined in atmi.h.    Also, looks like i need to add the reserved field here.

JonChesterfield:
ATMI has a lot of logic around hidden arguments that changes the value of kernarg segment size, terminating in:
```
// system.cpp
    info.kernel_segment_size =
        (hasHiddenArgs ? kernel_explicit_args_size : kernel_segment_size) +
        sizeof(atmi_implicit_args_t);
```

It's not clear to me how num_args * 8 matches this. Could you rig up a test case that uses your logic and the existing one and blows up noisily if they differ?

gregrodgers:
This works.   Merge it!   The code you quoted is stuff you will be deleting.   Like the code you quoted this adds  sizeof(atmi_implicit_args_t) to the numargs*8.    I believe there is other atmi code that resets kernel_segment_size to numargs*8.  But that happens later.  The fact that I am getting the same value as the comgr code gets is what makes this correct and allows you to get rid of the comgr code.    

I have another pr that does a lot of cleanup in hostcall.   And I am going to do lots more cleanup before I add new service call for vargs  host function pointer.   I don't want to pull rank and just merge this.   But the commits to plugins/hsa/impl are going to back up.   You will have peace after we carefully reduce the code so it is more understandable.   I also have to integrate a better hostrpc thread which will also reduce the code around hostcall.

ashwinma:
> ATMI has a lot of logic around hidden arguments that changes the value of kernarg segment size, terminating in:
> 
> ```
> // system.cpp
>     info.kernel_segment_size =
>         (hasHiddenArgs ? kernel_explicit_args_size : kernel_segment_size) +
>         sizeof(atmi_implicit_args_t);
> ```
> 
> It's not clear to me how num_args * 8 matches this. Could you rig up a test case that uses your logic and the existing one and blows up noisily if they differ?
> 
> Essentially the printfs you have commented out, but with a comparison that'll fail the smoke tests if they differ.

ATMI's logic is for all generic GCN kernel cases with each arg having different size/alignment. However for OpenMP/libomptarget the implementation is that each argument is 8 bytes (size and alignment). So it is num_args * 8 + sizeof(atmi_implicit_args). 

The interop function `atmi_interop_hsa_get_kernel_info` returns the size of only the non-implicit args by substracting the sizeof(atmi_implicit_args).

openmp/libomptarget/plugins/hsa/impl/system.cpp
```diff
@@ -1516,6 +1518,41 @@ hsa_status_t populate_InfoTables(hsa_executable_t executable,
...
+    if (found != std::string::npos) {
+      std::string kernelName = sym_name.substr(0, found);
+      if (KernelInfoTable[gpu].find(kernelName) == KernelInfoTable[gpu].end()) {
+        ErrorCheck(Finding expected kernel name in KernelInfoTable,
+                   HSA_STATUS_ERROR_INVALID_CODE_OBJECT);
+      }
+      atl_kernel_info_t kinfo = KernelInfoTable[gpu][kernelName];
+      atmi_kern_desc_t kdv;
+      atmi_status_t err = atmi_memcpy(&kdv, (void *)info.addr, sizeof(kdv));
```
JonChesterfield, line 1533: atmi_memcpy doesn't seem right here. Will write a bit more below

gregrodgers: The atmi_memcpy is because it is a d2h since the kernel is loaded and frozen.

JonChesterfield:
Thanks guys. Nearly with you now. If I have a handle on what atmi is doing, we can definitely bring the start up overhead down.

The *_kern_desc struct instance is emitted in the device code, same as before. It contains a bunch of openmp specific things - mode, maxParalleLevel etc. This patch adds the number of arguments to it, and because we currently only pass 8 byte aligned, 8 byte wide arguments that's enough to work out how much kernarg memory is needed. I don't like this because the data is already available in memory, but the redundancy isn't that bad.

The control flow in the plugin, as far as I can tell, is:
- load the device image (an elf64) into host memory
- 'freeze' it, which I believe is 'link it' then copy onto the gpu
- iterate over all symbols, storing them in a map
- for the kernel symbols, also look up the corresponding _kern_desc
- copy the bytes from the kern_desc back onto the host using atmi_memcpy
- use that information to preallocate memory etc

There are a couple of things here that seem suboptimal. One is iterating over & caching all the symbols in a map, when they're already available in a hash table in the elf. The second is copying the kern_desc onto the gpu, finding it on the gpu, then copying it back to the host in order to use it.

Proposals:
- The kern_desc data should be read on the host, as it's on the host to begin with. Ideally never copied onto the gpu, but failing that, at least not copied back. No atmi_memcpy needed. That's an existing inefficiency which this patch doesn't make worse.
- Iterating over all the symbols in an elf in order to build an in-memory representation of a map from symbol to address is redundant with the hash table already in the elf. We should only look up the symbols we need, and probably don't need the heap allocated cache.

Requires:
- A metadata parser, ideally added to hsa to implement the current API
- Checking hsa's get symbol by name uses the elf hash table (specifically not a linear scan) and fixing it if it doesn't
- Some more digging around in the elf on the host to avoid copying the kern_desc across and/or emitting it in the host code instead of the device

If we have said metadata parser to make startup quicker, we can also lift the number of arguments out of the metadata instead of redundantly storing it in kern_desc. Either with or without the simplifying assumption of 8 byte arguments.

PR #39, opened 2020-05-01
In this cleanup, I moved the service id header to a common place .  I…

gregrodgers:
… also got rid of one hostcall header.  I removed the hostcall_handler and called hostcall_assign_buffer directly.  There is lots more to cleanup with hostrpc


gregrodgers:
Thank you , I am going to merge this.   The registration of the handlers will go away in the next cleanup.    See my email about this.

PR #40, opened 2020-05-01
Message pack parser, proof of concept

JonChesterfield:
This is a clean slate implementation of a message pack parser, written from the spec.

The implementation is a bit rough and ready. I think it's better to get this up quickly, as a viable path off comgr, and then polish it in tree. For example, I haven't documented how the parser works. Some things are done at runtime that should be done at compile time. No allocation or > 8 byte copies anywhere in the parser and it is expected to gracefully handle corrupt input. Likely to be very fast post cleanup.

The plugin file system.cpp is modified to call into the proposed parser at every call site to comgr for code object v3, and any difference between the results is treated as a hard error. The comgr result is then discarded and the new (and equal) one used for the rest of the function. With this break-fast code in place, the tests still all pass locally.

The parser is written to cleanly expose the tree structure of the data format. This is wrapped in a layer to match the comgr calls for easier back to back comparison. That layer can largely disappear once the comgr comparison is removed.

Proposed sequence:
- Land this, and wait for a few days to check the new parser behaves identically to the comgr one for everyone else
- Replace v2 path with an error about unsupported object, remove redundant comgr calls for v3
- Comgr is now dead and is deleted from the plugin
- Refactor system.cpp to take advantage of the foreach-style parser api, instead of looking things up one at a time
- Optional, rework parser to specialise at compile time. Likely to generate better code.
- Background, continue testing parser error handling, API & implementation changes. E.g. namespace the symbols.

The parser itself will be easier to review post-cleanup. Provided it's reviewed before release I hope that's OK.

@ashwinma aomp is planning to error out on v2 objects. Does atmi need to support them? If not, the post-cleanup version of this may be useful for you. It handles variable size/align arguments exactly as atmi does.


gregrodgers:
Looks good.  Lots of renaming needed but that will come later I am sure.  I am merging this.

JonChesterfield:
Yeah, names are all over the place. Symbols aren't all local etc.

ashwinma:
@JonChesterfield ATMI does not need v2. LLC's default is v3 anyway. It was largely in place because AOMP's default was v2 for a while. 

A word of caution/warning: right now messagepack seems to work fine. If lightning changes to some other format in v4, you will have to redo this whole exercise. The ideal way is to push these changes to ROCr -- please don't give up on that path.

JonChesterfield:
Good news re: not needing v2. Do you have some test coverage for the non-8 byte argument case? I'd like to see that working back-to-back before dropping the comgr path.

Yeah, when/if llc changes from messagepack to something else we can repeat this exercise. That's OK though. I'll propose a patch to rocr once this stabilises. Unit tests are in https://github.com/JonChesterfield/msgpack as we don't have infra in place to run them.

PR #41, opened 2020-05-01
Drop support for code object v2

JonChesterfield:
Removes get_code_object_custom_metadata_v2

The get_code_object_version function checked the same notes section as find_metadata, so find_metadata only fails iff get_code_object_version would return != 3. Therefore dropped get_code_object_version by calling find_metadata before the comgr functions.

Also drops the unused atmi_platform_type_t platform from get_code_object_custom_metadata.

There are more parts of the v2/v3 abstraction that can now be folded. This patch is mainly intended to remove the substantial duplication between get_code_object_version and find_metadata introduced by #40.

### DIFF
--- a/openmp/libomptarget/plugins/hsa/impl/system.cpp
+++ b/openmp/libomptarget/plugins/hsa/impl/system.cpp
@@ -1031,176 +1031,6 @@ static amd_comgr_status_t populateCodeProps(
   return AMD_COMGR_STATUS_SUCCESS;
 }
 
-hsa_status_t get_code_object_custom_metadata_v2(atmi_platform_type_t platform,
-                                                void *binary, size_t binSize,
-                                                int gpu) {
-  amd_comgr_metadata_node_t programMD;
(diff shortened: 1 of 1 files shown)


openmp/libomptarget/plugins/hsa/impl/system.cpp
```diff
@@ -1836,10 +1600,12 @@ atmi_status_t Runtime::RegisterModuleFromMemory(void **modules,
     if (types[i] == AMDGCN) {
       // Some metadata info is not available through ROCr API, so use custom
       // code object metadata parsing to collect such metadata info
+
```
ashwinma, line 1603: unnecessary blank lines?

JonChesterfield: Probably a bunch of formatting slightly off here. I ran clang-format on the file but that generated a lot of spurious changes, using a different format file I guess. Inclined to fix it as we go along.

PR #43, opened 2020-05-05
Remove code related to builtins not needed, passes lots of tests

ronlieb:


ronlieb:
i reverted this change, I found 3 tests failing to compile , in spec accel, and hpc2020

JonChesterfield:
That's interesting. Which builtin was the error on?

PR #44, opened 2020-05-05
Remove use of comgr from hsa plugin

JonChesterfield:
Approximately lines up the debug error printing with that from comgr.


JonChesterfield:
Can't move openmp earlier in the build order because the deviceRTL looks for opencl device libraries which aren't necessarily available. The hip option `-nogpulib` should suppress the search, but that's currently broken. This patch leaves it the build order unchanged.

PR #49, opened 2020-05-06
[NFC] Msgpack. Start removing the type erasure overhead

JonChesterfield:
Introduce an alternative to the std::function interface to the parser, use it from skip_next_message and map_lookup_string. Removes the allocation associated with std::function and significantly cleans up codegen as a result.

Temporarily coexists with the std::function implementation to reduce the diff size.

### DIFF
--- a/openmp/libomptarget/plugins/hsa/impl/msgpack.h
+++ b/openmp/libomptarget/plugins/hsa/impl/msgpack.h
@@ -83,6 +83,72 @@ struct functors {
       };
 };
 
+template <typename Derived> class functors_defaults {
+public:
+  void cb_string(size_t N, const unsigned char *str) {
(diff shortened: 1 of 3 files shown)


openmp/libomptarget/plugins/hsa/impl/msgpack.h
```diff
@@ -83,6 +83,72 @@ struct functors {
       };
 };
 
+template <typename Derived> class functors_defaults {
+public:
+  void cb_string(size_t N, const unsigned char *str) {
```
JonChesterfield, line 88: The separation of interface and implementation (cb_foo vs handle_foo) may not be worthwhile. The useful part is being able to override (strictly, hide) array and still call the default array_elements or vice versa.

PR #51, opened 2020-05-06
Remove appending of libaompextras to pure HIP compile

ronlieb:
Error noticed by running hipcc without libompextras present. Should not need libaompextras for pure hip compile/build.

### DIFF
--- a/clang/lib/Driver/ToolChains/HIP.cpp
+++ b/clang/lib/Driver/ToolChains/HIP.cpp
@@ -487,8 +487,8 @@ void HIPToolChain::addClangTargetOptions(
     else
       WaveFrontSizeBC = "oclc_wavefrontsize64_off.amdgcn.bc";
 
-    BCLibs.append({"hip.amdgcn.bc", "opencl.amdgcn.bc", "ocml.amdgcn.bc",
-                   "ockl.amdgcn.bc", "oclc_finite_only_off.amdgcn.bc",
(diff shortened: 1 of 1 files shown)


clang/lib/Driver/ToolChains/HIP.cpp
```diff
@@ -487,25 +487,12 @@ void HIPToolChain::addClangTargetOptions(
     else
       WaveFrontSizeBC = "oclc_wavefrontsize64_off.amdgcn.bc";
 
-    // FIXME remove double link of aompextras and hip
-    if (DriverArgs.hasArg(options::OPT_cuda_device_only))
```
scchan, line 491: I think you could check whether the offload kind is OpenMP and then link in the additional extra library?

ronlieb: we have code elsewhere in the file that adds it for OpenMP, this appears to be left over.

ronlieb: Greg suggested I make the code match what is upstream, so I am testing that change now.

scchan: where that bitcode file is being added for OpenMP?  Isn't this the right place to do it?

ronlieb: const char *AMDGCN::Linker::constructOmpExtraCmds

PR #52, opened 2020-05-07
Revert back to using err_shared_var_init instead of warn_shared_init

ronlieb:
Clean build of aomp, and testing shows zero issues.  
Makes us match upstream.


ronlieb:
amusing at some level.

PR #53, opened 2020-05-07
Implement malloc as hostcall, proof of concept

JonChesterfield:
We shouldn't commit this as-is, but it's interesting that it passes our testing. At least locally. We may want to commit a cleaned up version which actually frees the memory.

The advantage over the current implementation is correctness when multiple kernels call malloc simultaneously, the disadvantage is it'll be much slower.

Long term we should:
- Remove all uses of malloc from the runtime, as we always have more context than 'some heap'
- Write a malloc which uses infrequent hostcalls to get large blocks of memory and carves it up on the device in thread safe fashion

### DIFF
--- a/openmp/libomptarget/deviceRTLs/amdgcn/src/memoryheap.cu
+++ b/openmp/libomptarget/deviceRTLs/amdgcn/src/memoryheap.cu
@@ -44,7 +44,12 @@
 DEVICE char gpuHeap[SIZE_OF_HEAP];
 DEVICE uint32_t gpuFlags[NUM_PAGES];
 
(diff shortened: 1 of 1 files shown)


ronlieb:
we have no tests that exercise tasks on target or data_sharing on target. Those are the only two remaining calls to SafeMalloc inside the runtimes for our target. Nexted parallel for was the agregious usage, which was resolved by, yep, allocating the storage on host and passing it in only when needed.

JonChesterfield:
I've belatedly read our malloc implementation. It's considerably more broken than I realised. I'll wire up free and propose a patch that just uses hostcall.

JonChesterfield:
Obsolete

PR #55, opened 2020-05-12
Use readcyclecounter instead of checking architecture version macro

JonChesterfield:
The AMDGCN macro is causing some merge problems internally. This change loses functionality, as readcyclecounter currently never expands to the more accurate memrealtime, but the loss seems acceptable.

The right implementation of wtime needs to check for the presence of the s-memrealtime feature, and choose a scaling factor which may be architecture-specific. This should probably be implemented in llc, not in the openmp library.


ronlieb:

Jon, it seems after applying this patch, that test/smoke/schedule fails.
Could you look into that failure?

JonChesterfield:
That doesn't seem right, but I can't look into it now. Reverted this patch.

JonChesterfield:
I think schedule fails with this because the timer overflows. It doesn't fail for me locally and I haven't put much time into trying to reproduce. I think we should reinstate this patch, as it removes the problematic macro, and someone can fix up schedule later.

PR #57, opened 2020-05-13
Cleanup dead code as found by gerrit reviews

ronlieb:
removal of dead code as show by testing all aomp tests.
These were pointed out during the gerrit code reviews.
Testing reveals code does not trigger failures.

### DIFF
--- a/clang/lib/Basic/Cuda.cpp
+++ b/clang/lib/Basic/Cuda.cpp
@@ -165,26 +165,6 @@ CudaVersion MaxVersionForCudaArch(CudaArch A) {
   case CudaArch::SM_20:
   case CudaArch::SM_21:
     return CudaVersion::CUDA_80;
-  case CudaArch::GFX600:
-  case CudaArch::GFX601:
-  case CudaArch::GFX700:
-  case CudaArch::GFX701:
-  case CudaArch::GFX702:
-  case CudaArch::GFX703:
-  case CudaArch::GFX704:
-  case CudaArch::GFX801:
-  case CudaArch::GFX802:
-  case CudaArch::GFX803:
-  case CudaArch::GFX810:
-  case CudaArch::GFX900:
-  case CudaArch::GFX902:
-  case CudaArch::GFX904:
-  case CudaArch::GFX906:
-  case CudaArch::GFX909:
-  case CudaArch::GFX1010:
-  case CudaArch::GFX1011:
-  case CudaArch::GFX1012:
-    return CudaVersion::CUDA_90;
   default:
     return CudaVersion::LATEST;
   }
--- a/clang/lib/CodeGen/CGOpenMPRuntime.cpp
+++ b/clang/lib/CodeGen/CGOpenMPRuntime.cpp
@@ -1675,6 +1675,11 @@ llvm::Value *CGOpenMPRuntime::emitUpdateLocation(CodeGenFunction &CGF,
   // OpenMPLocThreadIDMap may have null DebugLoc and non-null ThreadID, if
   // GetOpenMPThreadID was called before this routine.
   if (!LocValue.isValid()) {
+   // AMDGCN does not handle static initializers of aggregate constants.
+   // compiling -g will excercise this path.
+   if (CGM.getTriple().getArch() == llvm::Triple::amdgcn)
+      return getOrCreateDefaultLocation(Flags).getPointer();
+
     // Generate "ident_t .kmpc_loc.addr;"
     Address AI = CGF.CreateMemTemp(IdentQTy, ".kmpc_loc.addr");
     auto &Elem = OpenMPLocThreadIDMap.FindAndConstruct(CGF.CurFn);
--- a/clang/lib/CodeGen/CGOpenMPRuntime.cpp
+++ b/clang/lib/CodeGen/CGOpenMPRuntime.cpp
@@ -1612,15 +1612,6 @@ Address CGOpenMPRuntime::getOrCreateDefaultLocation(unsigned Flags) {
           CGM.GetAddrOfConstantCString(";unknown;unknown;0;0;;").getPointer();
       DefaultOpenMPPSource =
           llvm::ConstantExpr::getBitCast(DefaultOpenMPPSource, CGM.Int8PtrTy);
-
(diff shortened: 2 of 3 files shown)


JonChesterfield:
I'm not convinced these are all dead. Please hold off on committing for a little bit

clang/lib/CodeGen/CGOpenMPRuntime.cpp
```diff
@@ -9901,9 +9889,6 @@ void CGOpenMPRuntime::emitTargetNumIterationsCall(
     TD = getNestedDistributeDirective(CGM.getContext(), D);
   if (!TD)
     return;
-  const std::vector<llvm::Triple> &Devices = CGM.getLangOpts().OMPTargetTriples;
```
JonChesterfield, line 9904: I don't know the origins of this. Challenges navigating our periodically reinvented git history. It looks like an optimisation - avoids generating code when there are no devices.

clang/lib/CodeGen/CGOpenMPRuntime.cpp
```diff
@@ -1684,9 +1675,6 @@ llvm::Value *CGOpenMPRuntime::emitUpdateLocation(CodeGenFunction &CGF,
   // OpenMPLocThreadIDMap may have null DebugLoc and non-null ThreadID, if
   // GetOpenMPThreadID was called before this routine.
   if (!LocValue.isValid()) {
```
JonChesterfield, line 1677: Unless llc actully supports these now, I can't see how removing this code can be right

clang/lib/CodeGen/CGOpenMPRuntime.cpp
```diff
@@ -1612,15 +1612,6 @@ Address CGOpenMPRuntime::getOrCreateDefaultLocation(unsigned Flags) {
           CGM.GetAddrOfConstantCString(";unknown;unknown;0;0;;").getPointer();
       DefaultOpenMPPSource =
           llvm::ConstantExpr::getBitCast(DefaultOpenMPPSource, CGM.Int8PtrTy);
-
-      if (CGM.getTriple().getArch() == llvm::Triple::amdgcn) {
```
JonChesterfield, line 1616: llc still rejects the code that is generated here if we remove this patch. I'm not sure why our test suite no longer trips over it, but I don't think we should remove this.

clang/lib/Basic/Cuda.cpp
```diff
@@ -165,26 +165,6 @@ CudaVersion MaxVersionForCudaArch(CudaArch A) {
   case CudaArch::SM_20:
   case CudaArch::SM_21:
     return CudaVersion::CUDA_80;
-  case CudaArch::GFX600:
```
JonChesterfield, line 168: This is probably a fallback if no arch is specified. I think our tests always specific an arch though, so we wouldn't have any coverage of that

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
```diff
@@ -1246,13 +1246,10 @@ void CGOpenMPRuntimeNVPTX::GenerateMetaData(CodeGenModule &CGM,
   // Emitting Metadata for thread_limit causes an issue in ISEL, due to
   // an optimization in OPT.
   // See line 230 lib/Target/AMDGPU/AMDGPULowerKernelAttributes.cpp
-  bool enableMetaOptBug = false;
   bool flatAttrEmitted = false;
   int FlatAttr = 0;
   int DefaultWorkGroupSz =
       CGM.getTarget().getGridValue(GPU::GVIDX::GV_Default_WG_Size);
```
JonChesterfield, line 1252: I don't know what this meta opt bug is, so am personally fine with removing this. I can't see who wrote this from git log though so don't know who to check with.

clang/lib/Basic/Cuda.cpp
ronlieb: the comment from the reviewer is that an earlier  test of the arch causes this to be dead code, appears correct o me.

clang/lib/CodeGen/CGOpenMPRuntime.cpp
ronlieb: I too have a hard time finding history on a lot of these changes. Hopefully this is behind us now for changes going forward. I don't know what issue/test triggered this addition.

clang/lib/CodeGen/CGOpenMPRuntime.cpp
ronlieb: i put an assert here instead of the code, and trips all over the place in our smoke tests.  smoke/teams for example.

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
ronlieb: This was an issue with meta data we ran into in llvm 9, it now appears to be not an issue in llvm 11

clang/lib/Basic/Cuda.cpp
JonChesterfield: Oh, the IsAMDGpuArch? Yeah, fair enough. These functions probably shouldn't be called for AMD code anyway.

ronlieb:
Hypothetically speaking, all of these comments received on gerrit review would also have occurred upstream, and in either case would result in the changes being added to the commit into amd-stg-open and then these would just flow into the amd-stg-openmp branch via merge. 

Many of these original source additions have no associated test and are highly suspect as to whether they were needed at all.  Review questions motivated me to try and determine if the code was needed vai testing.

This specific PR does not need to land in amd-stg-openmp directly, it (or some variant) will arrive in the next month or so via the merge.  

The advantage to applying this PR is we get much better testing over time in our product sooner rather than later.

to quote someone "Pay me now, or pay me later."

ronlieb:
Greg ok'ed merging these changes, to help reduce upstream delta

JonChesterfield:
I'm sorry to hear that. The static initialization patch covers missing functionality in llc which is still missing. Hopefully it'll take less time to run to ground next time it surfaces.

PR #58, opened 2020-05-14
define macro to indicate old definition usage of s_memtime

ronlieb:


JonChesterfield:
Exposing whether the hardware has s-memrealtime as a macro seems reasonable. I'm not sure about the implementation though.

Can we use the feature map defined in clang/lib/Basic/Targets/AMDGPU.cpp? It has fields like `Features["s-memrealtime"] = true` which would map cleanly onto AMDGPU_HAS_FEATURE_S_MEMREALTIME or similar.

The naming should probably be 'has s-memrealtime' not 'use s-memtime' to match up with the above.

ronlieb:
discarding

PR #62, opened 2020-05-14
Fix an rpath related lit test, Jan Sjodin is the author

ronlieb:


ronlieb:
aomp specific I think.

PR #63, opened 2020-05-15
Disable malloc.

JonChesterfield:
Disable malloc on the grounds that it is broken beyond economic repair. 

No thread safety or bounds checks. If allocating < 64 bytes, attempts to write '1' to sixteen gigabytes of memory.

Our tests don't rely on malloc, so things will still look OK. Before the next release, we should use hostcall to implement malloc and free (see #53 for an initial sketch). That'll be slow but should actually work.


JonChesterfield:
What is 620.dash?

Please don't revert this. I'd much rather add the test to expected failing.

I'd expect to have the via-hostcall alternative running next week.

ronlieb:
620.dash, actually 620.dashmg is one of my SPEC HPG HPC2020 offload benchmarks, that passes. Adding it to known failures, is simply you telling me to remember its broken.
my life is simplified if we revert the patch, i view the disable malloc patch as a very successful experiment, it shows where do have a previously unknown dependence on malloc, and gives us a test to verify you host malloc patch.

JonChesterfield:
Can you implement malloc as a call to

`EXTERN char *  printf_alloc(uint32_t bufsz);`

as a stop gap instead? Leaking memory is unlikely to be an issue for the one test case

ronlieb:
yes, that works very nicely.
want me to push it up for review? or just push it in

JonChesterfield:
Cool. Reassuring that it worked. Should do really - printf is one of our more frequently exercised code paths. We should probably come up with a thorough test suite for memor allocation from the device.

I quite like having GitHub diffs as a place for discussion and/or reverting, even if the review itself isn't necessary.

PR #64, opened 2020-05-15
constrain deduplication workaround to AMDGCN only

ronlieb:


ronlieb:
I see a few issues in my SPEC offload testing with this patch, that I need to go look into.

ronlieb:
better now, less failures, I have one regression 620.dash which I need to look into.

Is there a better way to get at the target triple than what I did here?

gregrodgers:
I checked.  The upstream value is also false.  Can we change the entire file to be the same as upstream?

ronlieb:
We need this patch locally due to the correctness issues it fixes. This is the deduplication fix that jon created for the March merge

JonChesterfield:
Well, either this sort of patch or fix llc to lower alloca

ronlieb:
how about both!

Lets commit this patch as it matches what will get committed to amd-stg-open (eventaully).

Jon, would you open JIRA and assign to some llc person?

JonChesterfield:
I can have a go at Jira if you can point me to some docs on what the many fields need to be set to.

I suspect stack allocation already has a task somewhere, but don't know how to find it.

JonChesterfield:
I don't like the string compare, but also don't currently have any cycles to find the better way to do it. So feel free to commit, we can always fix up that later.

ronlieb:

How about if your write up the problem description and send it to Sai-Poorna to submit to JIRA on our behalf.

PR #66, opened 2020-05-16
patch4 review cleanup  gerrit 357444

ronlieb:
cleanup per review comments on gerrit , and verified with testing.


ronlieb:
Talked to greg, he OK'ed merging cleanup patch

PR #68, opened 2020-05-22
More cleanup from patch #4 review

ronlieb:


JonChesterfield:
Changes outside of CodeGenFunction.cpp look good.

Could you land those, but delay the commented part until Greg can take a look or we discuss it in one of the meetings?

JonChesterfield:
> promoting all pointers in kernarg to global, shouldn't be needed or done here
Sure, but that's consistent with HIP and Greg disagreeing about address spaces on arguments

PR #69, opened 2020-05-23
remove paramater processing for addrspace

ronlieb:
extracted one controversial change for further review .


JonChesterfield:
Thanks for extracting this. Greg?

JonChesterfield:
It's possible that the contention here is because openmp is passing kernel arguments in fine grain hsa memory, as the documentation requests, and hip is passing kernel arguments in gpu memory using architecture specific undocumented behaviour.

PR #71, opened 2020-05-24
patch6 cleanup

ronlieb:


ronlieb:
Our amd-stgg-openmp branch had previously turned off vectorization , I had removed that code as part of the merge to gerrit. I left those two comment lines in by mistake.

PR #73, opened 2020-05-26
more cleanup from gerrit reviews

ronlieb:
patch 4 and 6 cleanups


ronlieb:
I believe it is the upstream verison.

PR #74, opened 2020-05-28
consume  -mlink-builtin-bitcode in clang-build-select-link to make HI…

ronlieb:
…P.cpp code more consistent


gregrodgers:
I don't think so.   First question is why?

PR #75, opened 2020-05-30
More cleanup from review of patch 6, -std=c++11 and include

ronlieb:
Depends on related patch to fix test cases in aomp

### DIFF
--- a/llvm/lib/Support/CommandLine.cpp
+++ b/llvm/lib/Support/CommandLine.cpp
@@ -2493,7 +2493,7 @@ public:
 #ifdef PACKAGE_VENDOR
     OS << PACKAGE_VENDOR << " ";
 #else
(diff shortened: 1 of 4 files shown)


llvm/lib/Support/CommandLine.cpp
```diff
@@ -2493,7 +2493,7 @@ class VersionPrinter {
 #ifdef PACKAGE_VENDOR
     OS << PACKAGE_VENDOR << " ";
 #else
-    OS << "LLVM (http://llvm.org/):\n  ";
+    OS << "AOMP-11.5-1 (http://github.com/ROCm-Developer-Tools/aomp):\n Source ID:11.5-1-build_from_release_source\n  ";
```
JonChesterfield, line 2496: Probably don't want this part. Can we fix our patch system to get rid of this recurring annoyance?

ronlieb: it is indeed annoying, thanks for noticing.

perhaps Ethan could figure out how to remove this from our development builds?

PR #77, opened 2020-06-09
generalize the vargs processing in printf to work for any vargs funct…

gregrodgers:
…ion that we want to execute on the host, including a vargs function that uses a function pointer.  

Here is the implementation of  EmitHostrpcVargsFn which is now used by both printf and three host services. 1)hostrpc_varfn_uint, 2)hostrpc_varfn_uint64, and 3)hostrpc_varfn_double.   To see how these host functions are used as general purpose function pointers, see the new example in examples/openmp/host_varfn_function.


openmp/libomptarget/deviceRTLs/hostcall/src/hostcall_stubs.hip
```diff
@@ -13,13 +13,38 @@
...
 // generated by the clang compiler when it encounters a printf statement.
-// printf_alloc is implemented as a hostcall stub. We assume that the 
+// printf_allocate is implemented as a hostcall stub. We assume that the 
 // host routine for printf_execute will free the buffer that was allocated
-// by printf_alloc.
+// by printf_allocate.
 
-EXTERN char * printf_alloc(uint32_t bufsz) {
+// These 4 stubs all use the same MALLOC SERVICE routine
+EXTERN char * printf_allocate(uint32_t bufsz) {
```
JonChesterfield, line 23: These all do the same thing. Even if we're stuck with N symbols, they should all call the same helper function to do the work

openmp/libomptarget/deviceRTLs/hostcall/src/hostcall_stubs.hip
```diff
@@ -45,6 +70,36 @@ EXTERN int printf_execute(char * print_buffer, uint32_t bufsz) {
   return (int) result.arg0;
 }
 
+EXTERN uint32_t hostrpc_varfn_uint_execute(char * print_buffer, uint32_t bufsz) {
```
JonChesterfield, line 73: These are almost identical to each other, suggest a helper function that takes the service id as a parameter

openmp/libomptarget/plugins/hsa/impl/hostcall_handlers.c
```diff
@@ -29,20 +29,60 @@ SOFTWARE.
...
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 
 void handler_HOSTCALL_SERVICE_PRINTF(void *cbdata, uint32_t service, uint64_t *payload) {
     size_t bufsz          = (size_t) payload[0];
     char* device_buffer   = (char*) payload[1];
-    amd_hostcall_error_t err = hostcall_printf(device_buffer,bufsz);
-    payload[0] = (uint64_t) err;
+    uint uint_value;
```
JonChesterfield, line 43: Same duplication again, this time can pass in a function pointer. Slightly harder to parameterise on the memcpy, suggest each function memcpy to an integer of the same size as the type then zero extends.

openmp/libomptarget/plugins/hsa/impl/hostrpc.h
```diff
@@ -0,0 +1,103 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+*/
+
+// Need to get llvm typeID enum from Type.h but cant include LLVM headers in
+// runtime #include "../../../../../llvm/include/llvm/IR/Type.h" For now, we a
+// have a manual copy of llvm TypeID enum
+enum TypeID {
+  // PrimitiveTypes - make sure LastPrimitiveTyID stays up to date.
```
JonChesterfield, line 31: this looks likely to bite us. either a test that checks it matches llvm, or generate it from the llvm enum?

openmp/libomptarget/plugins/hsa/impl/hostrpc.h
```diff
@@ -0,0 +1,103 @@
...
+typedef struct hostrpc_pfIntRegs hostrpc_pfIntRegs_t; // size = 48 bytes
+
+struct hostrpc_pfRegSaveArea {
+  hostrpc_pfIntRegs_t iregs;
+  uint128_t freg[NUMFPREGS];
+};
+typedef struct hostrpc_pfRegSaveArea
+    hostrpc_pfRegSaveArea_t; // size = 304 bytes
+
+struct hostrpc_ValistExt {
```
JonChesterfield, line 86: why packed? why separate integer and floating point? I'm wondering if this is x86 specific

openmp/libomptarget/plugins/hsa/impl/hostrpc_varfn.c
```diff
@@ -0,0 +1,1051 @@
...
+// Add a floating point value when building va_list for vprintf
+static hostrpc_status_t hostrpc_pfAddFloat(hostrpc_ValistExt_t *valist,
+                                           char *numdata, size_t valsize,
+                                           size_t *stacksize) {
+  // FIXME, we can used load because doubles are now aligned
+  double dval;
+  if (valsize == 4) {
+    float fval;
+    memcpy(&fval, numdata, 4);
+    dval = (double)fval; // Extend single to double per abi
```
JonChesterfield, line 146: halfs get extended as well

openmp/libomptarget/plugins/hsa/impl/hostrpc_varfn.c
```diff
@@ -0,0 +1,1051 @@
...
+                                                uint32_t *rc) {
+  //
+  // Users are instructed that their first arg must be a dummy
+  // so that device interface is same as host interface. To match device
+  // interface we make the first arg be the function pointer.
+  //
+  hostrpc_varfn_uint_t *vfnptr = (hostrpc_varfn_uint_t *)fnptr;
+
+  switch (NumArgs) {
+  case 1:
```
JonChesterfield, line 478: lots of noise here. either macros, or #include a file with the contents?

openmp/libomptarget/deviceRTLs/hostcall/src/hostcall_stubs.hip
gregrodgers: Or I could change the frontend and make them all use printf_allocate.   See the change in CGExpr.cpp.  I have a fixme in there anyway.

gregrodgers:
Unless there are strong objections, lets squash and merge this and start the build for 11.6-1.

ronlieb:
[AMD Public Use]

It builds fine, just started some tests, should know preliminary results in 15 minutes

JonChesterfield:
This leaves us in a better place than we are today, so I'm good with merging it. I'd give it the tick in the gui but that's blocked by Ron's requested changes flag. Aside from the stylistic things, my remaining concerns are:

1/ this will magically handle char* in most cases, which will make it look like we handle it in all cases
2/ tight coupling with clang/llvm internal representation, which may change on us
3/ pretty complicated logic without any unit tests, I'm not totally confident this will be bugfree

2 & 3 are our debt. 1 is a bad if users depend on this feature and we later have to remove it on the grounds that it sometimes runs off into invalid memory, but I think we already have that with printf.

ronlieb:

Looks good to me, I will go approve it. Forgot I had requested a change.

PR #78, opened 2020-06-16
get rid of hostcall register functions.  Replace with handlePayload(s…

gregrodgers:
…ervice, payload).  handlPayload has a simple switch to call the correct service handler function.

Jon, This is the next step in simplifying hostcall into hostrpc.  Eventually, I want to merge atmi_hostcall.c hostcall.cpp, and hostrpc_handlers.c into a single source called hostrpc.c.   The only externs will be the few routines that atmi calls.

### DIFF
--- a/openmp/libomptarget/plugins/hsa/impl/hostcall.cpp
+++ b/openmp/libomptarget/plugins/hsa/impl/hostcall.cpp
@@ -125,11 +125,8 @@ struct record_t {
     bool discarded;
 };
 
-typedef std::function<void(uint64_t *)> service_handler_t;
-
 struct critical_data_t {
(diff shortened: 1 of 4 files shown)


openmp/libomptarget/plugins/hsa/impl/hostcall.cpp
```diff
@@ -194,6 +185,8 @@ get_payload(buffer_t *buffer, ulong ptr)
 
 static bool hostcall_version_checked;
 
+extern "C" void handlePayload(uint32_t service, uint64_t *payload);
```
JonChesterfield, line 188: How long is payload here? 8 or 64 seem likely, could write as 'uint64_t payload[8]' to make it more obvious

PR #79, opened 2020-06-16
cleanup nvptx macro

ronlieb:


JonChesterfield:
Is this another aomp-only patch that we can drop without anything breaking?

ronlieb:
yes

JonChesterfield:
Great!

PR #81, opened 2020-06-18
This moves the hostrpc stubs and host fallback functions into a singl…

gregrodgers:
…e library and include file. It also makes the stubs openmp cpp source instead of hip and reorganizes everything in the directory openmp/libomptarget/hostrpc


JonChesterfield:
The restructuring looks fine, I don't think there are any changes in behaviour here.

Checking in a bunch of IR is a bad thing though - because it wasn't hand written it's nearly illegible. I'd much rather us keep that in opencl

gregrodgers:
This PR now has a 2nd commit which replaces hostrpc_invoke.ll with hostrpc_invoke.cl

PR #82, opened 2020-06-18
Jc hostrpc

JonChesterfield:
This is incomplete, posting as a potential discussion point


JonChesterfield:
This collides with the cmake changes for the port to openmp, the source changes are OK. It essentially swaps out hostcall_invoke as was.

PR #84, opened 2020-06-23
Delete unused DataTask

JonChesterfield:
Dead code.


estewart08:
Pulled in the patch, rebuilt openmp, no unexpected errors with gfx906 testing.

JonChesterfield:
Thanks!

ashwinma:
Just FYI -- in case you want to add async data movement, you may want to revisit data tasks. I understand from a upstream point of view, it is a tad excessive at this point.

JonChesterfield:
Hey Ashwin. Yeah, there's a bunch of interesting runtime things that could be done here.

Optimising for upstream or the unified compiler in the near future means staying with the libomptarget plugin, which doesn't have the hooks to offload much to atmi. 

Lots of the local copy of atmi is unused as a result, my plan is to delete all the bits that aren't in use to unblock reviews.

ashwinma:
Yep, makes perfect sense to me.

PR #85, opened 2020-06-23
Ignore environment variable ATMI_DEPENDENCY_SYNC_TYPE, unused by openmp

JonChesterfield:
There's no reference to this environment variable outside of ATMI. Explicitly using the default is a step towards taskgroup being dead.

### DIFF
--- a/openmp/libomptarget/plugins/hsa/impl/data.cpp
+++ b/openmp/libomptarget/plugins/hsa/impl/data.cpp
@@ -171,8 +171,6 @@ atmi_status_t Runtime::Memfree(void *ptr) {
   return ret;
 }
 
-
-
 atmi_status_t Runtime::Memcpy(void *dest, const void *src, size_t size) {
   atmi_status_t ret;
   hsa_status_t err;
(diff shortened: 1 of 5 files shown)


JonChesterfield:
Added the fallout from removing the getenv:
- replace calls to getDepSyncType with the previously default value, as it's now the only returned value
- replace tests against that constant with true/false via find and replace
- delete the newly dead branches, leaving some braces in position to make the diff more legible

The same process could be applied to data.cpp, but #84 deletes the corresponding functions anyway.

ashwinma:
Just FYI -- this flag controls how the kernel dependencies are handled. Callback means that the background ROCr thread will wait for kernels to complete before launching the next kernel in chain of kernels. Barrier packet inserts HSA barrier packet in the AQL queue to manage kernel dependencies. While the variable is defaulted to callback in OpenMP, you can use barrier packets for better performance in case you want asynchronous task graphs to execute. Again, from the perspective of upstreaming, this may be of lower priority.

ashwinma:
Also taskgroups are one way in which you can implement streams in ATMI (taskgroup.ordered = true implies an in-order stream). One more FYI -- in case you want to revisit in the future.

estewart08:
Pulled in patch, rebuilt openmp, no expected failures on gfx906.

JonChesterfield:
Yep, execution is synchronous at present. Moving off that is going to be an interesting interaction between the openmp language, Johannes's work on the llvm runtime and how hip implements streams. I don't know what the solution will look like yet

PR #86, opened 2020-06-24
Remove most of cpu task and some other unused code

JonChesterfield:
OpenMP only initializes ATMI with devtype == GPU. Specialize on that then dead code elimination.

I thought hostcall was routed through the async cpu task, but on closer inspection one calls `process_packet` and the other `process_packets`. Hostcall uses threads that are independent from the ATMI cpu tasks.

A subset of ATLCPUProcessor remains as it's used in memory management for the GPU.


estewart08:
On build of amd_llvm_project I am seeing multiple failures in regards to enqueue_barrier_tasks. Some more cleanup may be necessary:
```
amd-llvm-project/openmp/libomptarget/plugins/hsa/impl/task.cpp:1464:11: error: ‘enqueue_barrier_tasks’ was not declared in this scope
```

JonChesterfield:
That's strange. With this diff, task.cpp ends at line 1368 (before your reported error line) and grep for enqueue_barrier_tasks across the whole tree comes up empty. I think this collides partly with the last couple of patches, so it might not apply cleanly without a git pull first.

estewart08:
Yeah, I was missing the changes from #85. Pulled in patch, rebuilt openmp, no unexpected failures on gfx906.

JonChesterfield:
Cool, thanks! I'll land this then. Leaves about 11k LOC remaining in the host plugin.

PR #88, opened 2020-06-25
Merge unbundle archive code into offload bundler.

jsjodin:


jsjodin:
Did anyone want to approve? This is the same code as in https://reviews.llvm.org/D80816. I was planning to get the driver changes done before the 8th and retire the clang-unbundle-archive tool.

PR #90, opened 2020-06-28
Reduce memory footprint of openmp runtime

ronlieb:
One field wihin the structure __kmpc_data_sharing_worker_slot_static
  char Data[DS_Worker_Warp_Slot_Size];
renamed:
  char HugeData[DS_SLOT_SIZE];

Another field also renamed:
  char MasterData[DS_SLOT_SIZE];

The first worker Data significantly contributes to the overall size of
global object omptarget_nvptx_device_State.
This field is used for environment variable data sharing, and seems to
rarely occur in practice.

Approach: reduce two staticaly allocated Data arrays to 4 elements.
when a request for more slots happens we will simply use
__kmpc_simple_malloc to acquire more memory.

Previously:  2,383,307,776
with patch:    236,381,184

### DIFF
--- a/openmp/libomptarget/deviceRTLs/common/omptarget.h
+++ b/openmp/libomptarget/deviceRTLs/common/omptarget.h
@@ -34,6 +34,8 @@
 #define BARRIER_COUNTER 0
 #define ORDERED_COUNTER 1
 
+#define DS_SLOT_SIZE 4
+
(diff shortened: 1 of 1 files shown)


ronlieb:
interesting questions, i misread the return 0 in the header, so i need to redo this.

ronlieb:
abandon

PR #93, opened 2020-07-01
Jc hostrpc openmp

JonChesterfield:
Work in progress.

Replaces the current hostcall implementation with a novel one.

There's more code here than necessary, and this patch doesn't delete the current hostcall. So the line count delta is not representative.


ronlieb:
is this one DOA now ?

JonChesterfield:
Not dead, but badly prempted by other work. I'm hoping some of the bugs in llc (addressing modes) will be fixed by the time I get back to it. May as well close the PR now though.

PR #101, opened 2020-07-16
Read kernel data from elf on host instead of from device

JonChesterfield:
The compiler writes several objects into the gpu image to pass information to the runtime, e.g. KernDescVal.

Currently these are copied to the GPU, HSA queried to find their location, atmi_memcpy called to copy the bytes back over to the host. Given these objects have no loader relocations applied, the same data is already available in the elf before it is loaded onto the GPU. 

This patch reads the values from the elf by parsing the .hash section. It exit(1)'s if the value from HSA differs, such that running this through our testing provides reasonable evidence that the parsing is right. It also runs the tests using the bytes read from the disk image instead of the loaded image. A later patch will drop this redundancy.

Further patches can use the faster (and more complicated) DT_GNU_HASH section instead. Some of the remaining atmi_memcpy calls can probably also be elided in similar fashion. It should ultimately be possible to drop the iteration over all symbols presently done during load and the associated runtime map.

Some of these objects can also be moved to a non-loadable section to slightly decrease memory footprint.

### DIFF
--- a/openmp/libomptarget/plugins/hsa/src/rtl.cpp
+++ b/openmp/libomptarget/plugins/hsa/src/rtl.cpp
@@ -20,6 +20,7 @@
 #include <ffi.h>
 #include <fstream>
 #include <iostream>
+#include <libelf.h>
 #include <list>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


JonChesterfield:
See https://github.com/ROCm-Developer-Tools/aomp/issues/87 for the solution to deepcode.

PR #102, opened 2020-07-17
remove hostrpc fallback functions from libomptarget.  They are alread…

gregrodgers:
…h in libhostrpc.a.  Also, to find them we need to add -lhostrpc to host linker not only for device linking but final link

### DIFF
--- a/clang/lib/Driver/ToolChains/CommonArgs.cpp
+++ b/clang/lib/Driver/ToolChains/CommonArgs.cpp
@@ -653,8 +653,7 @@ bool tools::addOpenMPRuntime(ArgStringList &CmdArgs, const ToolChain &TC,
       CmdArgs.push_back("-lrt");
 
   if (IsOffloadingHost) {
-    if (TC.getTriple().isAMDGCN())
-      CmdArgs.push_back("-lhostrpc");
+    CmdArgs.push_back("-lhostrpc");
     CmdArgs.push_back("-lomptarget");
   }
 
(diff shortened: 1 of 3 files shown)


clang/lib/Driver/ToolChains/CommonArgs.cpp
```diff
@@ -653,8 +653,7 @@ bool tools::addOpenMPRuntime(ArgStringList &CmdArgs, const ToolChain &TC,
       CmdArgs.push_back("-lrt");
 
   if (IsOffloadingHost) {
-    if (TC.getTriple().isAMDGCN())
-      CmdArgs.push_back("-lhostrpc");
+    CmdArgs.push_back("-lhostrpc");
```
ronlieb, line 656: This creates q requirement that libhostrpc.so exist in all builds , this broke everything which is why I moved it under isAMDGCN()

ronlieb:
maybe I should ask the simpler question, is this patch upstreamable with a -lhostrpc  requirement ?

JonChesterfield:
We definitely don't want copies of the fallback functions and it makes sense to me that they live in hostrpc.

I don't think hostrpc should be a dynamic library. It should be a static one, linked into the HSA plugin shared library. Symbols hidden.

Later, when we have it working for cuda, I think it should also be statically linked into the cuda plugin shared library.

That lets the fallback implementation diverge - e.g. call HSA or cuda to allocate - without runtime switching on architecture - while still sharing source code between the two where useful.

ronlieb:
I like the suggestion of statically linking it in.

PR #103, opened 2020-07-21
Drop redundant hsa calls, carefully preserving control flow

JonChesterfield:
Replaces atmi_interop_hsa_get_symbol_info with a function with a similar type that looks in the elf.

Drops check_loader_vs_reader, each block guarded by that is erased.

Several atmi_memcpy calls are no longer performed which should influence performance measurements.

### DIFF
--- a/openmp/libomptarget/plugins/hsa/src/rtl.cpp
+++ b/openmp/libomptarget/plugins/hsa/src/rtl.cpp
@@ -795,6 +795,20 @@ int get_symbol_info_without_loading(char *base, size_t img_size,
   }
   return 1;
 }
+
+atmi_status_t interop_get_symbol_info(char *base, size_t img_size,
+                                      const char *symname, void **var_addr,
+                                      uint32_t *var_size) {
+  symbol_info si;
+  int rc = get_symbol_info_without_loading(base, img_size, symname, &si);
+  if (rc == 0) {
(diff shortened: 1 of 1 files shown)


JonChesterfield:
I'm seeing some local failures that may be from this, investigating

ronlieb:

Enumerate failures ?

openmp/libomptarget/plugins/hsa/src/rtl.cpp
```diff
@@ -1117,8 +1065,9 @@ __tgt_target_table *__tgt_rtl_load_binary(int32_t device_id,
 
       void *WGSizePtr;
       uint32_t WGSize;
-      err = atmi_interop_hsa_get_symbol_info(place, WGSizeName, &WGSizePtr,
-                                             &WGSize);
+      interop_get_symbol_info((char *)image->ImageStart, img_size, WGSizeName,
```
JonChesterfield, line 1068: this doesn't look right, missing an err =

JonChesterfield:
Smoke tests fixed by the above err =, suspect aomp-tests is locally out of date.

JonChesterfield:
This patch is probably fine. Various openmp apps are failing for me locally with or without it.

JonChesterfield:
Tests passing locally after rebasing to latest.

PR #104, opened 2020-07-21
Only skip host file for archives.

jsjodin:


jsjodin:
This fixes the clang-offload-bundler failures.

PR #107, opened 2020-07-22
Elide an atmi call, drive by fixes

JonChesterfield:
Hostrpc services has some remaining hooks into atmi. This removes one of them.

Also replaces some references to uint64_t with value copy, adds a missing stdbool header. Can now use a more specific atmi header.

Remaining uses are of atmi_malloc and atmi_free (and the SERVICE_DEMO uses atmi_memcpy).

### DIFF
--- a/openmp/libomptarget/hostrpc/services/hostrpc_execute_service.c
+++ b/openmp/libomptarget/hostrpc/services/hostrpc_execute_service.c
@@ -32,6 +32,7 @@ SOFTWARE.
 #include "hostrpc_internal.h"
 #include <ctype.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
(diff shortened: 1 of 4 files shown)


JonChesterfield:
atmi_malloc/atmi_free are calls to hsa_amd_memory_pool with a hook to record metadata which is used by atmi_memcpy. It's plausible that we'll want another library to fill in gaps in hsa, accessible to hostrpc and to plugin/hsa.

I'm not sure what the optimal layout is here. The host fallback functions should probably be available when libomptarget doesn't dlopen the hsa plugin, which means they can't be in said plugin.

JonChesterfield:
Doesn't include atmi_interop_hsa anymore, replaced with atmi_runtime. I'm hopeful about removing atmi_memcpy, at which point we no longer need atmi_malloc or atmi_free and can go direct to hsa.

PR #109, opened 2020-07-22
Claim to be hip when compiling openmp for amdgcn. Similar to the cuda…

JonChesterfield:
…/openmp relationship. Haven't thought through the consequences. This patch is for discussion, not for committing.

Part of debugging why CUDA_ARCH is set for openmp compilation on amdgcn.

There are other places that need to be fixed up if going down this path. E.g. getLangOpts().HIP is used to set uniform-work-group-size to true, which we probably don't want on OpenMP.

I think OpenMP on nvptx essentially claims to be cuda. Hip also sets some of the cuda related options in clang. OpenMP on amdgcn can claim to be hip, and optionally also claim to be cuda.

This feels like debt - hip should not be implemented in terms of cuda, and I'm not confident openmp should be implemented in terms of either.


JonChesterfield:
No longer desired by Ron

PR #110, opened 2020-07-23
Lock signal pool, refactor mutex use in rtl to match

JonChesterfield:


JonChesterfield:
Somewhat heavy handed implementation - the queue gets used from several places, so it's wrapped in a class instead of adding lock/unlock calls around the uses.

The pool in rtl.cpp follows the same pattern so added a corresponding raii class there.

Gone with static initialization of the mutexes but could be persuaded to dynamically initialize instead. This way avoids a global constructor call, but not much in it.

Moves run_target_team_region_mutex into a function static as requested.

JonChesterfield:
Dropping in favour of #123, same patch applied to aomp12

PR #111, opened 2020-07-24
Jc variant libm

JonChesterfield:
This approximates a bring up of libm via variant for openmp on amdgcn. Some debugging code left in, posting the PR to avoid losing track of the WIP. This passes most, but not all of our tests. modf fails with a memory fault, one of the hip/openmp ones fails to compile.

powif has been implemented for hip in upstream.

The _OPENMP macro which doesn't work to distinguish target compilation has been replaced in upstream by _OPENMP_NVPTX, set manually. That's the right approach for amdgcn as well.

Converting the hip headers to be more easily used from C is largely at https://reviews.llvm.org/D84476

Plan is to wait for the merge from upstream to get through amd-stg-open and into amd-stg-openmp and cherry pick D84476, then reconstruct this patch on the result.

### DIFF
--- a/clang/lib/Driver/ToolChains/Clang.cpp
+++ b/clang/lib/Driver/ToolChains/Clang.cpp
@@ -1169,21 +1169,24 @@ void Clang::AddPreprocessingOptions(Compilation &C, const JobAction &JA,
 
   // If we are offloading to a target via OpenMP we need to include the
   // openmp_wrappers folder which contains alternative system headers.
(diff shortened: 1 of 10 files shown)


ronlieb:
i presume this is still a WIP, and should move to gerrit at some point ?

JonChesterfield:
This needs to be rewritten based on the changes imported from trunk. Will abandon this PR.

PR #113, opened 2020-07-27
Read call_struct_size from host instead of from device

JonChesterfield:


JonChesterfield:
Tests passing. The interop_hsa_get_symbol_info is already in use on hotter paths.

Removes the failure mode and overhead of a device to host memcpy.

PR #114, opened 2020-07-27
Copy device_environment data before loading onto gpu

JonChesterfield:
NFC. Removes an atmi_memcpy at image load time. Adds a hook between deserialize and load where the device image can be mutated by the caller, in this case to write the omptarget_device_environmentTy object.

Some line noise from routing through the C interface in atmi_runtime.h.

Previous patches to avoid atmi_memcpy (primarily because it launches a kernel on the gpu, which is not good for startup time) read data from the elf. This one writes to it.


JonChesterfield:
Closing in favour of #120, which is the same thing against aomp12

PR #116, opened 2020-07-28
Jc malloc expt

JonChesterfield:
Just got this running.

### DIFF
--- a/openmp/libomptarget/deviceRTLs/common/src/omp_data.cu
+++ b/openmp/libomptarget/deviceRTLs/common/src/omp_data.cu
@@ -28,9 +28,25 @@ DEVICE omptarget_device_environmentTy omptarget_device_environment;
 // global data holding OpenMP state information
 ////////////////////////////////////////////////////////////////////////////////
 
+#ifndef __AMDGCN__
+
 DEVICE
 omptarget_nvptx_Queue<omptarget_nvptx_ThreadPrivateContext, OMP_STATE_COUNT>
     omptarget_nvptx_device_State[MAX_SM];
+
+#else
+
+__attribute__((used))
+EXTERN uint64_t const constexpr omptarget_nvptx_device_State_size =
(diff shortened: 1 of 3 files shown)


JonChesterfield:
Concept proven. This patch is built on #114 which hasn't landed yet, but doesn't need to be. Will split them apart later today.

I'm seeing first launch latency change from 0.73s down to 0.11s. Ron reports a bigger change, may be something off in my local machine.

JonChesterfield:
Data structure.on GPU is identical to before. Same size, still all zeroed. This patch bypasses the slow handling of .bss in hsa's loader but makes no functional.change.

ronlieb:
is this patch intended  to be committed to aomp12 ?

JonChesterfield:
Waiting on feedback really. This might need to change if the structure becomes more dynamically sized. Do you think the implementation complexity is worth the latency reduction?

Longer term there's a path to making the loader faster, but equally we could then revert this.

ronlieb:
I think the latency improvement is in and of itself reasonable justification.
I think Singh would like to build on top of this to dynamically size the runtime footprint to the kernel(largest of the grou).
Think he can round up  the size to meet the 31 and 4 factors, are there others?

JonChesterfield:
OK, will sort this out then. The exact divisor trick was cute but fragile, it didn't survive the last merge from upstream. I'll implement the blocks-then-handle-tail logic. This version should build and run, but slower than it will once multiple CUs are used to do the memset once more.

JonChesterfield:
This runs (because it sets blocks = 1, which is always a factor) and is intended to be useful to Singh as-is. Rebased on recent aomp12.

The patch has some noise and parts of #114 still in the patch, will remove them early this week as part of getting this production ready.

JonChesterfield:
Simplified considerably by adopting hsa_amd_memory_fill. First launch taking about 0.1s locally, so the same ballpark as the custom kernel launch. Suggest we go with this.

JonChesterfield:
This is in the dev branch now

PR #118, opened 2020-07-30
[NFC} Update description of OMPGridValues enums

saiislam:
<Only for pre-phab review, not to be committed/merged in this repo>

Updated comments describing how OMPGridValues enums will be used in clang, deviceRTLs, and hsa and cuda plugins.

Please suggest if any other future use of these enums is not mentioned here.

### DIFF
--- a/llvm/include/llvm/Frontend/OpenMP/OMPGridValues.h
+++ b/llvm/include/llvm/Frontend/OpenMP/OMPGridValues.h
@@ -29,29 +29,30 @@ namespace omp {
 ///        use the new array name.
 ///
 /// Example usage in clang:
-///   const unsigned slot_size = ctx.GetTargetInfo().getGridValue(GV_Warp_Size);
+///   const unsigned slot_size = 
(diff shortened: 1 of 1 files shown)


gregrodgers:
I don't understand your first comment about not to be merged.   I like the comment update and would like this merged into AOMP.

saiislam:
> is this already in amd-stg-open branch of LC ?

No, I haven't committed it anywhere else.

saiislam:
> I don't understand your first comment about not to be merged. I like the comment update and would like this merged into AOMP.

My plan was to put it as an upstream patch first and then let it come to this branch. But, you can commit it, I will post it upstream as well.

ronlieb:
[AMD Public Use]

How about just putting it upstream first.

saiislam:
Here is the corresponding phab review: https://reviews.llvm.org/D86232

JonChesterfield:
The phab review is closed. amd-stg-open has presumably caught the update by now. Do we need to change aomp to match?

PR #119, opened 2020-08-01
11x merge and stg openmp

ronlieb:


ronlieb:
merged  from releases/11.x on July 31.

PR #125, opened 2020-08-05
Fix nogpulib in hip, lost part of upstream D68300

JonChesterfield:
Part of https://reviews.llvm.org/D68300, still present in amd-stg-open. Looks like a merge error.

Reintroducing this allows one to build the deviceRTL again (by passing nogpulib).

aomp-stg-openmp's HIP.cpp also ignores nogpulib, but I think we get a working build of deviceRTL anyway as the rocm search method is different. Suggest we add this to amd-stg-open anyway as users may expect nogpulib to work, especially when an error message recommends using it.


ronlieb:
Wael at ORNL ran into this problem building aomp 11, this patch needs to be cherry-picked to 11.x

PR #129, opened 2020-08-10
Add device libraries to HIP at target options phase.

jsjodin:


jsjodin:
Yes, we had a review open already with a different solution. I will update that one.

PR #132, opened 2020-08-14
Jc signals refactor

JonChesterfield:
The patch I most wanted is thread safe access to signals, but courtesy of the aomp12/aomp11 split the sanest way to pull the patch in pulls across some dead code deletion as well.

### DIFF
--- a/openmp/libomptarget/hostrpc/services/hostrpc_execute_service.c
+++ b/openmp/libomptarget/hostrpc/services/hostrpc_execute_service.c
@@ -227,11 +227,11 @@ static void hostrpc_handler_SERVICE_DEMO(uint64_t *payload) {
   int *A = (int *)malloc(N * sizeof(int));
   int *B = (int *)malloc(N * sizeof(int));
   int *C = (int *)malloc(N * sizeof(int));
-  copyerr = atmi_memcpy(A, A_D, N * sizeof(int));
-  copyerr = atmi_memcpy(B, B_D, N * sizeof(int));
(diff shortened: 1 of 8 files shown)


JonChesterfield:
I need this as the base to fix the race conditions in __tgt_rtl_data_submit and __tgt_rtl_data_receive. Will build that on top of this branch as it can't currently be flushed into amd-stg-openmp.

ronlieb:
when you get PR 132 done, am i better off trying to pull that into amd-stg-open rather than all the pieces that were in aomp12?
it seems like you are rolling lots into this PR to make it useful, which is fine by me.

JonChesterfield:
This removes a lot of the remaining races in the host runtime. aomp/issues/23 now passes, as does the parts of qmcpack I've succeeded in compiling.

Signal management is now thread safe. load_binary is locked as a whole, init_device is guarded by the caller. submit/receive data are now threadsafe. 

load_binary and run_target_region are still not safe to call simultaneously, which can arise on multigpu systems with multiple threads active.

JonChesterfield:
Landed the dead code elimination and signal pool rewrites, rebasing this on the result. Should be a much clearer diff

JonChesterfield:
> when you get PR 132 done, am i better off trying to pull that into amd-stg-open rather than all the pieces that were in aomp12?
> it seems like you are rolling lots into this PR to make it useful, which is fine by me.

I think so, yes. The plugin can and should be exactly the same in amd-stg-openmp, aomp12-as-was, amd-stg-open. And almost the same in trunk. I'd suggest we treat amd-stg-openmp as the authoritative source until the current release work is complete.

PR #136, opened 2020-08-15
Import async patches, setting up device environment without memcpy

JonChesterfield:
Comments plus eliding a memcpy. Leaves amd-stg-openmp and aomp12 plugins functionally equivalent.


JonChesterfield:
Device env part was at pull/120

PR #139, opened 2020-08-17
Reduce device RTL memory footprint

pdhaliwal-amd:


clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
```diff
@@ -1449,10 +1449,35 @@ void CGOpenMPRuntimeNVPTX::emitNonSPMDEntryHeader(CodeGenFunction &CGF,
   CGF.EmitRuntimeCall(
       createNVPTXRuntimeFunction(OMPRTL_NVPTX__kmpc_kernel_init), Args);
 
+  StringRef DataSharingMemorySlotName = "openmp.data.sharing.memory.slot";
+  size_t WarpSlotSize =
+      CGF.getTarget().getGridValue(llvm::omp::GV_Warp_Slot_Size);
+  size_t DataSharingMemorySlotSize = WarpSlotSize * 64;
+
+  // creating a global array which will be used for data sharing slots
```
JonChesterfield, line 1457: Why construct this in codegen instead of as an array in the devicertl?

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
```diff
@@ -1567,8 +1592,33 @@ void CGOpenMPRuntimeNVPTX::emitSPMDEntryHeader(
...
+    size_t WarpSlotSize =
+        CGF.getTarget().getGridValue(llvm::omp::GV_Warp_Slot_Size);
+    size_t DataSharingMemorySlotSize = WarpSlotSize * 64;
+
+    // creating a global array which will be used for data sharing slots
+    // This will be optimized in clang-build-select-link
+    llvm::Type *Ty =
+        llvm::ArrayType::get(CGF.CGM.Int8Ty,
+                           DataSharingMemorySlotSize);
+    llvm::GlobalVariable *DataSharingMemorySlot = new llvm::GlobalVariable(
```
JonChesterfield, line 1605: Is this the same array as above?

openmp/libomptarget/deviceRTLs/interface.h
```diff
@@ -433,8 +433,8 @@ EXTERN void __kmpc_kernel_prepare_parallel(void *WorkFn);
 EXTERN bool __kmpc_kernel_parallel(void **WorkFn);
 EXTERN void __kmpc_kernel_end_parallel();
 
-EXTERN void __kmpc_data_sharing_init_stack();
-EXTERN void __kmpc_data_sharing_init_stack_spmd();
+EXTERN void __kmpc_data_sharing_init_stack(char *Data, size_t size);
```
JonChesterfield, line 436: Probably can't modify these prototypes without also modifying nvptx, can add more functions

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
```diff
@@ -1449,10 +1449,35 @@ void CGOpenMPRuntimeNVPTX::emitNonSPMDEntryHeader(CodeGenFunction &CGF,
   CGF.EmitRuntimeCall(
       createNVPTXRuntimeFunction(OMPRTL_NVPTX__kmpc_kernel_init), Args);
 
+  StringRef DataSharingMemorySlotName = "openmp.data.sharing.memory.slot";
+  size_t WarpSlotSize =
+      CGF.getTarget().getGridValue(llvm::omp::GV_Warp_Slot_Size);
+  size_t DataSharingMemorySlotSize = WarpSlotSize * 64;
```
JonChesterfield, line 1455: This looks like a constant - I thought the idea was to compute this per-kernel?

pdhaliwal-amd: This is bit different approach than computing per kernel. For now, I am allocating 1MB of memory globally and uniquely per kernel. So the memory footprint is proportional to the number of non-spmd kernels. This size can be adjusted later during clang-build-select-link phase.

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
pdhaliwal-amd: because deviceRTL would not know about number of kernels present in the device image.

clang/lib/CodeGen/CGOpenMPRuntimeNVPTX.cpp
pdhaliwal-amd: Yes

openmp/libomptarget/deviceRTLs/interface.h
pdhaliwal-amd: Will do. Though these changes should work with nvptx as well but I don't know how to test.

pdhaliwal-amd:
@JonChesterfield I am first allocating 1MB of memory per kernel. And each kernel will have its own memory array. So, device memory footprint is directly dependent on the size of number of non-SPMD kernels. For e.g. a program having only one (non-spmd) kernel would take only 1MB of memory. The device_state var is now only around ~236MB from previous ~2.3GB.

JonChesterfield:
Per kernel or per launch? If per kernel, and this is the same underlying structure, I'd guess that's safe, though neither will be able to allocate the full 1mb and the second one to start may get none.

Memory footprint would be dependent on the total number of non-spmd kernels, not on the total number that are currently running, right? So we'll have a lot of memory that is unused. It seem that a test case with a lot of kernels would exceed the fixed 2.3gb we currently use

pdhaliwal-amd:
Yes, this is per kernel (emitting one global array per kernel).

It seem possible to use private memory in case where memory usage is low.

> It seems desirable to only occupy memory on the gpu for the kernels that are running, as then we can handle multiple instances of the same kernel correctly and minimise the footprint.
You are right. These changes do not take care of multiple instances of same kernels launched in parallel. Previous implementation has proper locks in place in case of large number of kernels.

PR #141, opened 2020-08-17
Use single device_state allocation per device

JonChesterfield:
Each host shared library containing amdgpu offload region contains it's own copy of the deviceRTL.

Prior to this patch, each copy was allocated a device_state object, currently about 2gb. This remains allocated until the end of program execution. Therefore a relatively small number of shared libraries exhausts GPU memory.

This patch allocates the storage once and uses the same data between multiple host shared libraries. It's a thread safe structure used by multiple kernels already so execution remains valid. We could improve this by exposing the size of the structure at compile time, so allowing allocation during init instead of on the first load_binary.

Treats multiple binaries with different size structures as a hard error. This assumption will fail when multiple architectures are supported from a single binary. Hopefully by then this structure will have been replaced with something more efficient.

### DIFF
--- a/openmp/libomptarget/plugins/hsa/src/rtl.cpp
+++ b/openmp/libomptarget/plugins/hsa/src/rtl.cpp
@@ -336,13 +336,15 @@ public:
   // Resource pools
   SignalPoolT FreeSignalPool;
 
-  // Clean up manually allocated state
   struct atmiFreePtrDeletor {
     void operator()(void *p) {
       atmi_free(p);  // ignore failure to free      
     }
   };
-  std::vector<std::unique_ptr<void, atmiFreePtrDeletor>> deviceStateStore;
+
+  // device_State shared across loaded binaries, error if inconsistent size
+  std::vector<std::pair<std::unique_ptr<void, atmiFreePtrDeletor>, uint64_t>>
+      deviceStateStore;
 
   static const int HardTeamLimit = 1 << 20; // 1 Meg
   static const int DefaultNumTeams = 128;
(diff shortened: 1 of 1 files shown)


JonChesterfield:
This is per-host-process. Fork is broken in a variety of ways already.

So if you run ./helloworld.exe, each will get it's own 2gb array.

If you spawn four threads and they call call the helloworld kernel, they'll all use the same array.

What this patch changes is when one host executable runs openmp code from N host shared libraries. Before they each got their own array, until memory ran out. With this patch, they all use the same array.

ronlieb:

Do you have a nice test case for smoke on this issue?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

JonChesterfield:
The devito tti benchmark aborts with OUT_OF_RESOURCES without this patch. No smoke test written that checks the behaviour change, did this with printf of allocated addresses and matching them to the corresponding free.

A smoke test would be one where N libraries each containing a target region are all loaded and executed sequentially, where N * 2gb > size of gpu memory, and the result is out of memory. That's probably reasonable to add.

Related, I think multiple library handling is broken on nvptx and amdgcn - load_binary discards a bunch of state related to previous binaries, but I can't see why that would be safe. The kernels from previous binaries can still be called.

JonChesterfield:
Added a test at aomp/pull/129, but actually clang falls over trying to build the first shared library. So that's currently a test for a bug in clang. Once that's fixed, it can become a test for this. I'm not sure how devito is building shared libraries, interesting that it's working for them.

PR #147, opened 2020-08-25
Refer to missing rocm instead of missing cuda in amdgpu openmp

JonChesterfield:


JonChesterfield:
new diag needs an argument that the previous one didn't, expressed as a crash. fixing

PR #149, opened 2020-08-27
Modify the gdb plugin to use Python 3 and other changes from ompd-tests

jinisusan:
Changes to modify the gdb plugin to use Python 3 (and other changes from the ompd-tests branch). I tested this with the rocm3.7 rocgdb (Python 3 based).


ronlieb (approved):
lgtm, built and tested on 11.x line

PR #150, opened 2020-08-28
Enforce better looping on sync to ensure kmpc_amd_master_end holds un…

gregrodgers:
…til worker is done

This may fix many problems we have with generic mode.   Nvidia uses a double partial barrier.  We use a full barrier.  But because printf or any user code could generate a barrier our full barriers need to reenter a barrier if the "other side" is not done.  For master the "other side" are workers.  For worker threads, the "other side" is the master thread.   This PR  sets the volatile attribute  boolean control variables omptarget_master_active and omptarget_workers_active because the "other side" is setting it.  I want to refrain from atomics here for performance reasons.  This PR fixes printf_parallel_for_target but may cause serious regression.

### DIFF
--- a/openmp/libomptarget/deviceRTLs/amdgcn/src/amdgcn_sync.hip
+++ b/openmp/libomptarget/deviceRTLs/amdgcn/src/amdgcn_sync.hip
@@ -27,6 +27,7 @@
 /// barrier.
 EXTERN void __kmpc_amd_worker_start(kmp_Ident *loc_ref, int32_t tid) {
   PRINT0(LD_SYNC, "call kmpc_amd_worker_start\n");
+  omptarget_master_active = true;
   omptarget_workers_active = false;
   __kmpc_impl_syncthreads();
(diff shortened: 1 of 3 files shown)


JonChesterfield:
I've been thinking about this. All-wave barriers don't compose very well. If the code in the worker waves calls barrier, it'll affect the master wave, which may never hit that point in the control flow. I think it would be possible to construct a test case where reasonable worker code deadlocks if run with an extra wave tagged on that does its own thing.

There seem to be two strategies for fixing this. One is to use a master dispatch, instead of a master wave. That prevents communication through LDS. Also prevents using barriers in the implementation. We could use a piece of GDS, though that may be removed in later architectures, or some global memory.

The other is to implement a partial barrier on top of atomics. That'll be slower than the hardware-accelerated barrier but easy to get right. It's probably equivalent to a refactoring of what we have now with the barriers removed.

Will continue thinking.

PR #151, opened 2020-09-01
Change two booleans to a single shared variable to manage generic sta…

gregrodgers:
…te. But the real fix is the removal of the barriers in the device stub functions in hostrpc


gregrodgers:
Using PR 152 instead

JonChesterfield:
Note that 152 also needs those barriers removed. They are removed on amd-stg-openmp, as they were never necessary, but this probably means an omp barrier in application code will fail if run as a generic kernel.

PR #152, opened 2020-09-01
Implement partial barrier

JonChesterfield:
Implements a partial barrier. Uses it to replace the amdgpu specific codegen for sync.


JonChesterfield:
This is an improvement on what we have now, in that it simplifies the codegen and unbreaks printf_parallel without regressing other tests. It needs a target-specific hook to initialize an integer to zero which I'll propose for upstream, with a no-op for nvptx.

The idea is to implement a counting barrier in 32 bits of LDS. Each wave increases a 16 bit counter until one reaches the total number we're waiting for, at which point that last wave is responsible for resetting the counter and releasing the waiting waves.

This is believed correct for the current call site - it is called with 'all the worker waves', which is the number in flight - 1. It may be prudent to handle the case where the number of waves reaching the barrier simultaneously can exceed the number that we are waiting for, which will probably cost a compare and swap.

Also, we may need some thread fences around this - the barrier as a whole should probably block loads/stores from migrating over it. That isn't caught by the current testing.

Finally this hasn't been directly unit tested and is a naive implementation.

JonChesterfield:
The fences have system scope. That is too broad - a cross wave barrier can probably restrict that to reduce overhead. I suggest that refinement be postponed for now in case that overhead is masking other latent bugs.

estewart08:
Tested with gfx906, reduction regressions are fixed. However, the printf smoke test still will occasionally seg fault.

JonChesterfield:
When debugging an earlier version of this, I sometimes saw the GPU printf passed a null buffer and fault. Also the hostrpc version check occasionally failed, one side reporting 0 and the other 6. Didn't get a clean repro. Suspect memory corruption, but I believe that to be preexisting. I sometimes have to run smoke tests a second time to get a pass.

estewart08:
One update, I saw this seg fault back on 11.6-2 as well, where the output was correct. So it is similar behavior to what we are seeing now. Definitely better than the incorrect output issues seen starting with 11.7-0.

JonChesterfield:
I think the segfault discussed above was caused by the hostrpc init function being called from multiple threads. Was patched by locking around init.

PR #154, opened 2020-09-01
find_package for Python3 does not update PYTHON_EXECUTABLE.

estewart08:
Instead it modifies Python3_EXECUTABLE. This is required as ompd support recently moved to python3.


JonChesterfield:
Yep, this unblocks me. Thanks, merging

PR #155, opened 2020-09-03
[OpenMP][AMDGPU] Use DS_Max_Warp_Number instead of WARPSIZE

pdhaliwal-amd:
This is the one of the main issue why memory usage on AMDGPU (~2.3GB) was
larger than NVIDIA (~671MB). With this change, the memory usage by
deviceRTL is around 770MB. This also reduces kernel launch latency.

### DIFF
--- a/openmp/libomptarget/deviceRTLs/common/omptarget.h
+++ b/openmp/libomptarget/deviceRTLs/common/omptarget.h
@@ -273,7 +273,7 @@ private:
       workDescrForActiveParallel; // one, ONLY for the active par
 
   ALIGN(16)
-  __kmpc_data_sharing_worker_slot_static worker_rootS[WARPSIZE];
+  __kmpc_data_sharing_worker_slot_static worker_rootS[DS_Max_Warp_Number];
   ALIGN(16) __kmpc_data_sharing_master_slot_static master_rootS[1];
 };
 


pdhaliwal-amd:
Note: this does not fix the smoke/data_share2 test.

ronlieb (approved):

JonChesterfield:
I read through the code and can't immediately tell whether this is correct. The control flow in the area is complicated. Could you note down how you worked this out?

gregrodgers:
Excellent catch!  It is a coincidence that WARPSIZE=Max_Warp_number = 32 for Nvidia.  Both NVIDIA and AMD have a max thread/workgroup size of 1024 threads.  For Nvidia 32*32=1024.   For amdgcn you can only have a maximum of 16 warps of size 64 in 1024 threads.     So DS_Max_warp_number=16.  I approve.

gregrodgers (approved):

gregrodgers:
Note: There is a grid value from OMPGridValues.h but Jon has not yet adopted the use of OMPGridValues.h in deviceRTL.  If he did, the value would be GRIDVAL[GV_Max_Warp_Number].   Since that is constant it should be able to size a static array.

estewart08:
Tested with gfx906: SOLLVE, smoke, examples, openmpapps and omp5 showed no regressions.

ronlieb:

Merge it says I

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

JonChesterfield:
Yep, pretty sure this is correct. Took me a while to walk through the control flow. Thanks!

PR #161, opened 2020-09-17
Adjust plugin cmake to closer match upstream

JonChesterfield:
Roughly corresponds to D87841 / gerrit 414960


JonChesterfield:
Can drop the llvm find_package after aomp/pull/150 lands

estewart08 (approved):

PR #162, opened 2020-09-18
lock around assign_buffer

JonChesterfield:


JonChesterfield:
Landed on amd-stg-open, cherry picking across to amd-stg-openmp to avoid occasional segv in prinf tests

PR #164, opened 2020-09-21
RFC: [OpenMP] Use shared memory for initial team

pdhaliwal-amd:


JonChesterfield:
Hey. What's the motivation for this change? LDS is quite scarce.

I'm not sure an openmp team always maps to a single CU. If gpu threads in use exceed approx 64*8 they can't use the same shared memory, and iirc we have tests that use >>512. That must be split over multiple kernels, and I don't know whether that is compatible with this change.

pdhaliwal-amd:
This is one of the few cases where global memory slots are being used. This patch is the one of many patches which aim towards removing memory slot infra from device rtl.

I haven't found any such case where shared memory, in this particular case, would introduce any regression. There is check on if enough LDS is available, only then it is going to use it otherwise the old logic still exists. I am going to change the old logic as well. But I first need to know if it is good to use shared memory in such cases or not. I know that this is going to affect occupancy as well. But I don't have that much experience to say whether it is bad or not.

ronlieb:
a team is constrained to one CU. Yes, LDS is super scarce. how did you come up with 64*8 as the limit ?

pdhaliwal-amd:
I have always found that a single team is mapped to single CU, the way barrier works, right?

JonChesterfield:
> a team is constrained to one CU. Yes, LDS is super scarce. how did you come up with 64*8 as the limit ?

That's good to know, thanks. 64*8 would be wavefront size * number of waves running at the same time on gfx9 (strictly 10, but I think it's usually limited to 8). So number of threads.

LDS would be 64k per CU, so 64/8 for 8k per wave.

pdhaliwal-amd:
Closing this

PR #165, opened 2020-09-24
[hostrpc] Fix wrong argument passed to atmi_free

pdhaliwal-amd:
The device free() was failing because wrong argument was being checked out on the host.


pdhaliwal-amd:
This fixes the smoke/data_share2 test case

ronlieb (approved):

JonChesterfield:
That's not ideal. Good catch, thanks

PR #166, opened 2020-10-02
Fix race condition in hostrpc cmake.

estewart08:
On occasion the libhostrpc archive is empty. This occured because the conversion
of the semicolon separated list to a space separated string was not ocurring before
the add_custom_command for the archive creation was invoked. Cmake actually
knows to ignore the semicolons in the list, which allows us to pass the list
directly into the command. Passed in CI.


JonChesterfield (approved):
Thanks!

PR #167, opened 2020-10-02
Jc openmp devicertl

JonChesterfield:
Not to land, creating PR to view diff in browser.


JonChesterfield:
Implemented, shipped.

PR #168, opened 2020-10-06
Embed version number into abs symbol for hostcall_invoke detection

JonChesterfield:
This change includes the hostcall version macros in hostrpc_invoke then embeds them in the symbol used to detect whether hostcall is in use.

Uses an absolute symbol to save the four bytes. Unfortunately, our loader segfaults on any absolute symbols while loading the executable. So that needs to be fixed before this lands.

### DIFF
--- a/openmp/libomptarget/deviceRTLs/amdgcn/src/hostrpc_invoke.cl
+++ b/openmp/libomptarget/deviceRTLs/amdgcn/src/hostrpc_invoke.cl
@@ -1,5 +1,7 @@
 #include "ockl_hsa.h"
 
+#include "../../../hostrpc/src/hostrpc.h"
+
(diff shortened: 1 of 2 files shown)


JonChesterfield:
Could use
```
// file scope, so can't detect whether hostrpc_invoke has been used with it
__attribute__((weak))
__attribute__((visibility("default")))
unsigned __hostcall_version = HOSTRPC_VRM;
```
or similar to embed the version number in the device runtime until the abs bug is fixed. The intent to do the version checking once instead of during each individual call.

JonChesterfield:
Reported the crash on weak in the assembler at https://bugs.llvm.org/show_bug.cgi?id=43187 but haven't looked into fixing it

PR #173, opened 2020-10-14
Rename plugin/hsa to plugin/amdgpu

JonChesterfield:
Leaving file updates to a second patch


JonChesterfield:
As amd-llvm-project/pull/172, but applied to aomp11 and with Ethan's report patched

estewart08 (approved):
Tested on gfx906, no unexpected failures.

PR #174, opened 2020-10-14
Call synchronous hsa_memory_copy

JonChesterfield:
Minimal patch. Faster in most, possibly all test cases. Exposes a bug in printf with variable format string, pfspecifier_str fails.

### DIFF
--- a/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
+++ b/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
@@ -355,6 +355,9 @@ public:
 
   atmi_status_t freesignalpool_memcpy(void *dest, const void *src,
                                       size_t size) {
+    hsa_status_t rc = hsa_memory_copy(dest, src, size);
+    return (rc == HSA_STATUS_SUCCESS) ? ATMI_STATUS_SUCCESS : ATMI_STATUS_ERROR;
+    
     hsa_signal_t s = FreeSignalPool.pop();
     if (s.handle == 0) {
       return ATMI_STATUS_ERROR;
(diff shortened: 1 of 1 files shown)


ronlieb (changes requested):
applied the patch to aomp12,  i am seeing some wrong answer issues on device.
i will probably need to spend time in the morning help triage this one ...

SPEC ACCEL 553.pclvrleaf and 556.psp fail as follows:

Libomptarget error: Copying data to device failed.
Libomptarget error: Call to targetDataBegin failed, abort target.
Libomptarget error: Failed to process data before launching the kernel.
Libomptarget error: run with env LIBOMPTARGET_INFO>1 to dump host-targetpointer
maps

similar issue with 605.lbm and 619.clvleaf from spec hpc2021 kit 26

JonChesterfield:
Error isn't in printf. Fails in roct.

`ioctl(3, AMDKFD_IOC_ALLOC_MEMORY_OF_GPU, 0x7ffe3150b4e0) = -1 EPERM (Operation not permitted)`

Reasonably sure the 23 bytes of data it is trying to copy is near the start of .rodata. HsaMemFlags passed HostAccess but otherwise clear, additionally setting ReadOnly doesn't help.

JonChesterfield:
Updated patch with the obvious work around. Try hsa_memory_copy, as it seems to be faster when it works, then fall back on the current logic if it doesn't. Possible stop gap until we can pinpoint why hsa_memory_copy fails.

ronlieb:
the revised patch seems to have resolved the the runtime errors in the four benchmarks i mentioned earlier.
i looked at the copy back time for my MI-Teams.cpp test case in ~/git/aomp-test/openmpapps/MI-Teams/src
and i do see an improvement from 
Time 5.022684 for gpu copy from device
to
Time 3.197447 for gpu copy from device

this is copying back aprox 5GB of data.

JonChesterfield:
I think we should land this. It moves the needle on performance, and is sufficiently obviously an ugly workaround that we're unlikely to forget about it. Should work up a small test case to send to the hsa guys, or debug through kfd to work out what the permission problem is, but that doesn't need to block this.

ronlieb (approved):

PR #178, opened 2020-10-16
[DeviceRTL][AMDGPU] Split atmi_memcpy for d2h and h2d

pdhaliwal-amd:
This patch is mostly NFC as the underlying copy is still done by generic atmi_memcpy. Call sites, however, are updated to use the respective copy method as they already know which direction they are going to copy in.  Future patch will remove the check for the direction from the memcpy implementation.

### DIFF
--- a/openmp/libomptarget/plugins/amdgpu/impl/atmi.cpp
+++ b/openmp/libomptarget/plugins/amdgpu/impl/atmi.cpp
@@ -35,7 +35,26 @@ atmi_status_t atmi_module_register_from_memory_to_place(
  */
 atmi_status_t atmi_memcpy(hsa_signal_t sig, void *dest, const void *src,
                           size_t size) {
+  hsa_status_t rc = hsa_memory_copy(dest, src, size);
+
+  // hsa_memory_copy sometimes fails in situations where
+  // allocate + copy succeeds. Looks like it might be related to
+  // locking part of a read only segment. Fall back for now.
+  if (rc == HSA_STATUS_SUCCESS) {
+    return ATMI_STATUS_SUCCESS;
+  }
+
   return core::Runtime::Memcpy(sig, dest, src, size);
+}
+
+atmi_status_t atmi_memcpy_h2d(hsa_signal_t sig, void *device_dest,
+                              const void *host_src, size_t size) {
+  return atmi_memcpy(sig, device_dest, host_src, size);
+}
+
+atmi_status_t atmi_memcpy_d2h(hsa_signal_t sig, void *host_dest,
+                              const void *device_src, size_t size) {
+  return atmi_memcpy(sig, host_dest, device_src, size);
 }
 
 atmi_status_t atmi_free(void *ptr) { return core::Runtime::Memfree(ptr); }
--- a/openmp/libomptarget/plugins/amdgpu/impl/atmi_runtime.h
+++ b/openmp/libomptarget/plugins/amdgpu/impl/atmi_runtime.h
@@ -155,53 +155,16 @@ atmi_status_t atmi_malloc(void **ptr, size_t size, atmi_mem_place_t place);
  */
 atmi_status_t atmi_free(void *ptr);
 
-/**
- * @brief Syncrhonously copy memory from the source to destination memory
- * locations.
- *
- * @detail This function assumes that the source and destination regions are
- * non-overlapping. The runtime determines the memory place of the source and
- * the
- * destination and executes the appropriate optimized data movement methodology.
- *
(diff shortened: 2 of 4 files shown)


openmp/libomptarget/plugins/amdgpu/impl/atmi.cpp
```diff
@@ -38,6 +38,16 @@ atmi_status_t atmi_memcpy(hsa_signal_t sig, void *dest, const void *src,
   return core::Runtime::Memcpy(sig, dest, src, size);
 }
 
+atmi_status_t atmi_memcpy_h2d(hsa_signal_t sig, void *host_dest,
```
JonChesterfield, line 41: parameter names are backwards. h2d as in 'host to device' copies from host to device, so host_src and device_dst. Though src/dst seems fine too.

openmp/libomptarget/plugins/amdgpu/impl/atmi_runtime.h
```diff
@@ -182,6 +182,16 @@ atmi_status_t atmi_free(void *ptr);
 atmi_status_t atmi_memcpy(hsa_signal_t sig, void *dest, const void *src,
```
JonChesterfield, line 182: drop atmi_memcpy from header, helps catch call sites that have been missed from the patch

ronlieb (approved):

openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
```diff
@@ -353,23 +353,42 @@ class RTLDeviceInfoTy {
   static const int Default_WG_Size =
       llvm::omp::AMDGPUGpuGridValues[llvm::omp::GVIDX::GV_Default_WG_Size];
 
-  atmi_status_t freesignalpool_memcpy(void *dest, const void *src,
-                                      size_t size) {
+  atmi_status_t freesignalpool_memcpy_h2d(void *dest, const void *src,
+                                          size_t size) {
```
JonChesterfield, line 357: The hsa_memory_copy hack should be in atmi_memcpy, as then it's in one place instead of four.

The two implementations are the same, other than the atmi_memcpy_ call, so can implement as one function that takes a pointer. Typing directly into this comment box so syntax may be slightly off:
```
template <typename F> // or use an explicitly typed function pointer
static atmi_status_t freesignalpool_memcpy_impl(void *dest, const void *src, size_t size, F func) {	

hsa_signal_t s = FreeSignalPool.pop();
if (s.handle == 0) {
  return ATMI_STATUS_ERROR;
}
atmi_status_t r = func(s, dest, src, size);
FreeSignalPool.push(s);
return r;
}

atmi_status_t freesignalpool_memcpy_h2d(void *dest, const void *src, size_t size) {	
  return freesignalpool_memcpy_impl(dest, src, size, atmi_memcpy_h2d);
}
atmi_status_t freesignalpool_memcpy_d2h(void *dest, const void *src, size_t size) {	
  return freesignalpool_memcpy_impl(dest, src, size, atmi_memcpy_d2h);
}
```

JonChesterfield (changes requested):
Broadly reasonable, though more code than necessary.

If you look at the start of Runtime::Memcpy, which atmi_memcpy wraps, you'll find the current logic for guessing the direction of copy. We can use that to check the directions are consistent, but that may as well be a second patch.

pdhaliwal-amd:
Updated with requested changes. I had seen atmi_memcpy_no_signal being used only in one place hostrpc_handler_SERVICE_DEMO method, so I had moved its definition for hostrpc_execute_service.

openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
```diff
@@ -353,27 +354,31 @@ class RTLDeviceInfoTy {
   static const int Default_WG_Size =
       llvm::omp::AMDGPUGpuGridValues[llvm::omp::GVIDX::GV_Default_WG_Size];
 
+  using MemcpyFunc = std::function<atmi_status_t (hsa_signal_t,
```
JonChesterfield, line 357: Generally in favour of std::function, though it does induce some runtime overhead. As part of making this library exception proof (which is some way off), I'd prefer we go with the raw function pointer `using MemcpyFunc = atmi_status_t (*)(hsa_signal_t, void*,const void*, size_t)` or similar.

JonChesterfield (changes requested):

JonChesterfield (approved):

JonChesterfield:
LG, thanks. Please squash the three commits into one before committing.

PR #181, opened 2020-10-20
Inlining logic for atmi_memcpy_{h2d|d2h}

pdhaliwal-amd:
NFC again. This time the logic for verifying the direction is there, so the map is still there. In the next patch, hopefully, the g_data_map should be gone. I had moved quite a code from data.cpp to atmi.cpp. I am unable to figure out how would I handle hostrpc DEMO logic so I had commented out the non working code. Why? Because I don't know how to get agent from given that much context. I checked other places like malloc and found by default device_id 0 being used.


openmp/libomptarget/hostrpc/services/hostrpc_execute_service.c
```diff
@@ -252,11 +253,11 @@ static void hostrpc_handler_SERVICE_DEMO(uint64_t *payload) {
   int *A = (int *)malloc(N * sizeof(int));
   int *B = (int *)malloc(N * sizeof(int));
   int *C = (int *)malloc(N * sizeof(int));
-  copyerr = atmi_memcpy_no_signal(A, A_D, N * sizeof(int), false);
-  copyerr = atmi_memcpy_no_signal(B, B_D, N * sizeof(int), false);
+//  copyerr = atmi_memcpy_no_signal(A, A_D, N * sizeof(int), false);
```
JonChesterfield, line 256: Don't want to break the demo like this, but we do indeed not have enough context to know which gpu is live. Hostcall doesn't work for multi-gpu systems at present. Therefore assume device 0 in this call.

JonChesterfield:
LGTM. Do the print/aborts trigger in any of our tests? Smoke / apps etc

JonChesterfield (approved):

PR #182, opened 2020-10-20
This fixes a bug in InstructionSimplify pass where a compare of two c…

gregrodgers:
…onstants does not have the same size.  By extending to the same bitwidth, the problem is resolved in the optimization pass

This PR fixes the smoke test reduction_array_section.   The problem is that we IR generation to initialize a statically sized array is creating a compare of two unlike constant sizes.  Some 32 and some 64.  They need to be the same type (and bitwidth).  This shows up for amdgcn because local address space pointers are 32 bit, whereas in nvptx, they are all 64 bit .

### DIFF
--- a/llvm/lib/Analysis/InstructionSimplify.cpp
+++ b/llvm/lib/Analysis/InstructionSimplify.cpp
@@ -698,12 +698,17 @@ static Constant *stripAndComputeConstantOffsets(const DataLayout &DL, Value *&V,
   assert(V->getType()->isPtrOrPtrVectorTy());
 
   Type *IntIdxTy = DL.getIndexType(V->getType())->getScalarType();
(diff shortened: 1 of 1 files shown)


ronlieb (commented):
does this change need to go into upstream llvm ?

JonChesterfield:
That's interesting. Seems likely to be a bug that affects any language targeting amdgcn, not just openmp. Change looks fine to me, but think this should probably be propagated to amd-stg-open or trunk too.

JonChesterfield (approved):
Spent a few minutes wondering about sext vs zext, then noticed this was pointer comparison.

gregrodgers:
Yes, this needs to be prepared for upstream review.  As discussed, all harding fixes in aomp11 need to go upstream.

PR #183, opened 2020-10-21
[libomptarget] Remove g_data_map

pdhaliwal-amd:
Checks added in the previous patch are now removed from atmi_memcpy and thus corresponding dead code.


JonChesterfield:
Good, but can go further. ATLPointerTracker is dead. Also, look at atmi_malloc and atmi_free, both of which contain stuff that is now dead.

JonChesterfield:
All good, thanks. Squashed locally, merging.

PR #189, opened 2020-10-22
change linkage on _exec_mode and _wg_size global to avoid duplicate s…

gregrodgers:
…ymbols such as kokkos is experiencing


JonChesterfield (approved):
Great! An extended version of this change should remove a bunch of if (amdgcn()) noise from this file

PR #190, opened 2020-10-22
Pr allow host offloading without opts

gregrodgers:


gregrodgers:
Sorry, this should only have the one commit.  I need to delete this PR and start over.

PR #192, opened 2020-10-22
Drop externally initialized qualifier on shared variables

JonChesterfield:


JonChesterfield:
Close approximation to gerrit 430245.

JonChesterfield:
Gerrit made it through.

PR #194, opened 2020-10-26
Pr fix floor ambiguous

gregrodgers:


clang/lib/Headers/__clang_hip_math.h
```diff
@@ -1338,6 +1338,18 @@ __DEVICE__ int min(int __arg1, int __arg2) {
 __DEVICE__ int max(int __arg1, int __arg2) {
   return (__arg1 > __arg2) ? __arg1 : __arg2;
 }
+__DEVICE__ unsigned int min(unsigned int __arg1, unsigned int __arg2) {
```
JonChesterfield, line 1344: Is this sufficient?
int x = -3;
unsigned y = 4;
return min(x, y);

PR #196, opened 2020-10-29
Update shared variable linkage. From D90248, G434046

JonChesterfield:
Trunk D90248 changed nvptx to use weakanylinkage/internal and undef variables, gerrit 434046 changed amdgcn to use the same. This patch updates aomp11 to match.

Primary purpose is removing the if (amdgcn) {some global} else {similar global} logic so that the same code works for nvptx and for amdgcn, picking from the intersection of the back end capability.

This patch includes some drive by whitespace cleanups and a branch around InitTempAlloca that looks like a merge error.


estewart08:
tested on gfx906, lgtm

estewart08 (approved):

PR #197, opened 2020-10-30
Update shared variable linkage, without breaking gfx803

JonChesterfield:
Reimplement pull/196, excluding the parts that break on gfx803. Left comments inline to that effect.

### DIFF
--- a/clang/lib/CodeGen/CGOpenMPRuntimeGPU.cpp
+++ b/clang/lib/CodeGen/CGOpenMPRuntimeGPU.cpp
@@ -7,7 +7,7 @@
 //===----------------------------------------------------------------------===//
 //
 // This provides a generalized class for OpenMP runtime code generation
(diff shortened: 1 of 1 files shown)


ronlieb (commented):
not clear to me, what did you have to do for this patch to work with gfx803 ?

JonChesterfield:
It does work for gfx803. At least, if I build it on r3, it passes the tests I expect it to.

ronlieb (commented):
i assumed it was working, i was wondering what needed to change and perhaps why gfx803 was not working before?

JonChesterfield:
Unclear - the private/internal linkage should make no difference, and the address space qualifier should make no difference either.

The bug could be in the loader, back end, hardware. Given no debugger and a symptom of 'invalid address' on code that works on later ISA's it's hard to pin down what the problem is.

ronlieb (approved):

PR #199, opened 2020-11-06
[hostrpc] Make malloc work with multiple devices

JonChesterfield:


JonChesterfield:
Ah, can't review this if I make the PR myself.

PR #200, opened 2020-11-06
[hostrpc] Make malloc work with multiple devices

pdhaliwal-amd:


openmp/libomptarget/hostrpc/services/hostrpc_externs.c
```diff
@@ -38,6 +38,7 @@ struct atl_hcq_element_s {
   buffer_t *hcb;
   hsa_queue_t *hsa_q;
   atl_hcq_element_t *next_ptr;
+  uint32_t device_id;
```
JonChesterfield, line 41: This field needs to be dropped, and the device_id field in hcb written to instead

JonChesterfield:
Landed in aomp11, with modifications

PR #203, opened 2020-11-16
add unintrusive trace wrapper

JonChesterfield:
API here is to #include a header in the plugin to toggle on debug printing.

### DIFF
--- a/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
+++ b/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
@@ -44,6 +44,8 @@ extern "C" hsa_status_t hostrpc_terminate();
 
 #include "omptargetplugin.h"
 
+#include "trace.h"
+
 // Get static gpu grid values from clang target-specific constants managed
 // in the clang header file GpuGridValues.h
 #include "llvm/Frontend/OpenMP/OMPGridValues.h"
--- a/openmp/libomptarget/plugins/amdgpu/src/trace.h
+++ b/openmp/libomptarget/plugins/amdgpu/src/trace.h
@@ -1,0 +1,334 @@
+#ifndef LIBOMPTARGET_PLUGINS_AMGGPU_SRC_TRACE_H_INCLUDED
+#define LIBOMPTARGET_PLUGINS_AMGGPU_SRC_TRACE_H_INCLUDED
+
+#include <array>
+#include <chrono>
+#include <cinttypes>
+#include <cstdio>
+#include <omptarget.h>
+#include <tuple>
+#include <utility>
+
+extern int print_kernel_trace; // set by environment variable
+
+namespace {
+namespace detail {
+
+// Plumbing for concatenating format strings
+template <size_t N, size_t... Is>
+constexpr std::array<const char, N - 1> toArray(const char (&a)[N],
+                                                std::index_sequence<Is...>) {
+  return {{a[Is]...}};
+}
+template <size_t N, size_t... Is>
+constexpr std::array<const char, N - 1> toArray(const char (&a)[N]) {
+  return toArray(a, std::make_index_sequence<N - 1>());
+}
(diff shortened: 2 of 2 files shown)


JonChesterfield:
The mechanism may not be obvious. For each external function in the plugin,
1/ declare foo_impl
2/ write foo_impl() { return foo(); }
3/ rename foo() to foo_impl

The foo_impl functions should probably be static, though it doesn't matter much.

My suggestion is to check in trace.h, leaving it as dead code. Then when someone wants to switch this on while debugging amdgcn or nvptx plugins, `#include "trace.h"` and rebuild.

JonChesterfield:
Now checks the environment variable as requested. Adds ~100 lines to fold the N printf calls per function into one. Generated code for an eight argument run_target is as follows, looks good enough to me. There's a negligible cost to the (predictable) branch and code size increase, but it's within the measurement noise.

```
@__func__.__tgt_rtl_run_target_team_region = hidden unnamed_addr constant [33 x i8] c"__tgt_rtl_run_target_team_region\00", align 1
@_ZN12_GLOBAL__N_16detail6fmtStrIJRiRPvRPS3_RPlS2_S2_S2_RmEE5valueE = hidden constant %"struct.std::array.107" { [82 x i8] c"Call %35s: %8ld ms (%14d, 0x%.12lx, 0x%.12lx, 0x%.12lx, %14d, %14d, %14d, %14lu)\0A\00" }, align 1

define dso_local i32 @__tgt_rtl_run_target_team_region(i32 %0, i8* %1, i8** %2, i64* %3, i32 %4, i32 %5, i32 %6, i64 %7) local_unnamed_addr #0 {
  %9 = load i32, i32* @print_kernel_trace, align 4, !tbaa !2, !noalias !6
  %10 = icmp eq i32 %9, 2
  br i1 %10, label %13, label %11

11:                                               ; preds = %8
  %12 = tail call i32 @_ZL37__tgt_rtl_run_target_team_region_impliPvPS_Pliiim(i32 %0, i8* %1, i8** %2, i64* %3, i32 %4, i32 %5, i32 %6, i64 %7) #4
  br label %20

13:                                               ; preds = %8
  %14 = tail call i64 @_ZNSt6chrono3_V212system_clock3nowEv() #4, !noalias !6
  %15 = tail call i32 @_ZL37__tgt_rtl_run_target_team_region_impliPvPS_Pliiim(i32 %0, i8* %1, i8** %2, i64* %3, i32 %4, i32 %5, i32 %6, i64 %7) #4
  %16 = tail call i64 @_ZNSt6chrono3_V212system_clock3nowEv() #4
  %17 = sub nsw i64 %16, %14
  %18 = sdiv i64 %17, 1000
  %19 = tail call i32 (i8*, ...) @printf(i8* nonnull dereferenceable(1) getelementptr inbounds (%"struct.std::array.107", %"struct.std::array.107"* @_ZN12_GLOBAL__N_16detail6fmtStrIJRiRPvRPS3_RPlS2_S2_S2_RmEE5valueE, i64 0, i32 0, i64 0), i8* getelementptr inbounds ([33 x i8], [33 x i8]* @__func__.__tgt_rtl_run_target_team_region, i64 0, i64 0), i64 %18, i32 %0, i8* %1, i8** %2, i64* %3, i32 %4, i32 %5, i32 %6, i64 %7) #4
  br label %20

20:                                               ; preds = %13, %11
  %21 = phi i32 [ %12, %11 ], [ %15, %13 ]
  ret i32 %21
}
```

A slightly simpler implementation spuriously copied the ~100 byte string to the stack before calling printf. It is tempting to fold __func__ into the format string as well, which works, but probably isn't strictly permitted by c++. Minor saving relative to cutting out the extra printf calls, which itself is borderline whether the runtime improvement justifies the code complexity.

JonChesterfield:
Squashed the commits, good to go from my perspective.

JonChesterfield:
This is working convincingly for debugging devito, and if it causes problems it's easy to disable by dropping the include.

PR #204, opened 2020-11-18
Limit default number of teams to number of CUs

pdhaliwal-amd:
Default number of teams when launched with #pragma omp target teams would be equal to the number of CUs. Previously it was hard coded to 128.

### DIFF
--- a/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
+++ b/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
@@ -754,9 +754,16 @@ int32_t __tgt_rtl_init_device(int device_id) {
     DP("Default number of teams set according to environment %d\n",
        DeviceInfo.EnvNumTeams);
   } else {
-    // default number of teams is 1:1 with number of compute units.
-    DeviceInfo.NumTeams[device_id] = DeviceInfo.ComputeUnits[device_id];
-    DP("Default number of teams set according to number of compute units %d\n",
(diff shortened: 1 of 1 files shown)


ronlieb (commented):
have you compared performance before and afer this change with any of our openmpapps?
or other performance tests ?

pdhaliwal-amd:
Not yet. There is one aomp_mappings smoke which is failing right now. Most probably its checking the hard coded value of teams as 128. Since now that is changed, I think that test will need to be updated as well.

ronlieb:
regarding aomp_mappings test, its going to produce different output for number of teams as the range of gpus we see have varying number of CUs for different models.  i have seen numbers from 12 to 120. so we might want to add a num_teams caluse to make the test behave more uniformly

JonChesterfield:
`default number of teams is 1:1 with number of compute units` <- sounds reasonable to me

gregrodgers (changes requested):
change the environment variable name to OMP_DEFAULT_TEAMS_PER_PROC   .  This same value can be used for other architectures and for host teams.   If this works out for us, we will want to put in a spec change request for OpenMP 5.2.

gregrodgers:
Per teams discussion, the name should be OMP_TARGET_TEAMS_PER_PROC.  Maybe in another pull request we can get an api called omp_get_device_num_procs(int devid) that would return the number of CUs or SMs on the device.

gregrodgers (approved):

PR #205, opened 2020-11-18
print return value from traced tgt calls

JonChesterfield:
Prints in format `Call               __tgt_rtl_load_binary:     3050ms 0x0000017754a0 (             0, 0x000000409a40)`, i.e. result then arguments, as that lines up with how the functions are declared and means the result values fall in the same column.

### DIFF
--- a/openmp/libomptarget/plugins/amdgpu/src/trace.h
+++ b/openmp/libomptarget/plugins/amdgpu/src/trace.h
@@ -99,31 +99,32 @@ template <typename... Ts> constexpr auto fmtTuple() {
 // are the same type, worked around using the longer spelling.
 // Writing the contents of fmtStr::get() inline in log_t is simpler, but puts
 // a ~100 byte object on the stack and calls memcpy on it.
-template <typename... Ts> class fmtStr {
+template <typename R, typename... Ts> class fmtStr {
   static constexpr auto get() {
-    // Call function: time ms (some, number, of, arguments)
-    return cat(toArray("Call %35s: %8" PRId64 " ms "), fmtTuple<Ts...>(),
-               toArray("\n\0"));
+    // Call function: 123ms result (some, number, of, arguments)
+    return cat(cat(toArray("Call %35s: %8" PRId64 "ms "),
+                   fmt<typename std::decay<R>::type>::value(), toArray(" ")),
+               cat(fmtTuple<Ts...>(), toArray("\n\0")));
   }
 
 public:
   static constexpr size_t size() { return get().size(); }
-  static constexpr const std::array<const char, fmtStr<Ts...>::size()> value =
-      get();
+  static constexpr const std::array<const char, fmtStr<R, Ts...>::size()>
+      value = get();
   static constexpr const char *data() { return value.data(); }
 };
-template <typename... Ts>
-constexpr const std::array<const char, fmtStr<Ts...>::size()>
-    fmtStr<Ts...>::value;
+template <typename R, typename... Ts>
+constexpr const std::array<const char, fmtStr<R, Ts...>::size()>
+    fmtStr<R, Ts...>::value;
 
-template <typename... Ts> struct log_t {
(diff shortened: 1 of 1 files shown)


ronlieb:
lgtm in general. when do we want to tie in the LIBOMPTARGET_KERNEL_TRACE control ?  
or do we need to at all, since the feature needs to be explicitly included to enable, correct?

i would also comment that i can achieve the same without a patch
 
 stdbuf --output=0 ltrace  -tt -e __tgt\* ./helloworld

JonChesterfield:
ltrace is interesting. It doesn't seem to be crossing the dlopen boundary, is that configurable?

ltrace:
```
15:55:09.655698 helloworld->__tgt_register_requires(1, 0x7ffc816ba738, 0x7ffc816ba748, 256)                                      = 1
15:55:09.655920 helloworld->__tgt_register_lib(0x40aea0, 0x7ffc816ba738, 0x7ffc816ba748, 256)                                    = 0
15:55:09.851171 helloworld->__tgt_target(-1, 0x40205e, 1, 0x7ffc816ba640Hello world. 100
Hello world. iteration 0
Hello world. iteration 1
Hello world. iteration 2
Hello world. iteration 3
Hello world. iteration 4
)                                                        = 0
Target region executed on the device
15:55:09.867310 helloworld->__tgt_unregister_lib(0x40aea0, 0x40cdc0, 0x40cdc8, 0)                                                = 0
```

diy:
```
Call         __tgt_rtl_number_of_devices:        0ms              2 )
Call           __tgt_rtl_is_valid_binary:        1ms              1 (0x00000040ae80)
Call             __tgt_rtl_init_requires:        0ms              1 (             1)
Call               __tgt_rtl_init_device:       32ms              0 (             0)
Call               __tgt_rtl_load_binary:     3132ms 0x000000b6b420 (             0, 0x00000040ae80)
Call                __tgt_rtl_data_alloc:        1ms 0x7fc168a0a000 (             0,              4, 0x7ffecd1abd44)
Call         __tgt_rtl_data_submit_async:       87ms              0 (             0, 0x7fc168a0a000, 0x7ffecd1abd44,              4, 0x7ffecd1abc08)
Hello world. 100
Hello world. iteration 0
Hello world. iteration 1
Hello world. iteration 2
Hello world. iteration 3
Hello world. iteration 4
Call   __tgt_rtl_run_target_region_async:     3170ms              0 (             0, 0x000000abb6d0, 0x000000b74dc0, 0x000000b70720,              1, 0x7ffecd1abc08)
Call       __tgt_rtl_data_retrieve_async:       99ms              0 (             0, 0x7ffecd1abd44, 0x7fc168a0a000,              4, 0x7ffecd1abc08)
Call               __tgt_rtl_data_delete:        1ms              0 (             0, 0x7fc168a0a000)
Call               __tgt_rtl_synchronize:        0ms              0 (             0, 0x7ffecd1abc08)
Target region executed on the device
Call           __tgt_rtl_is_valid_binary:        8ms              1 (0x00000040ae80)
```

ronlieb:
remove the -e __tgt*    option, as i added that so as to only see __tgt calls.
then you will see more calls

ronlieb:
have you considered adding the include of this into the lib-debug build ?
then we can enable with LIBOMPTARGET_KERNLE_TRACE anytime we like.
and of course we can always explicitly include in none debug build when desired.

gregrodgers:
We need this tracing all the time regardless of the use of lib-debug.   The change was already made to see this when LIBOMPTARGET_KERNEL_TRACE=2.

ronlieb:
ok, i see it thx

ronlieb (approved):

JonChesterfield:
If we can replace this with ltrace + grep, that's great. I haven't been able to get ltrace to look through dlopen. Google suggests this is a known limitation that has been fixed in newer versions of ltrace. I'm unclear if the dlsym/pointer table logic in libomptarget interferes, or if my ltrace is just too old.

PR #206, opened 2020-11-23
Remove unused field in kern_desc

JonChesterfield:
WG_Size and Mode look redundant to me, leaving them in place for now


ronlieb (approved):

JonChesterfield:
version and TSize are pretty close to unused too, so this can probably be simplified to just maxparallellevel

ronlieb:
agreed, they can vanish by the wayside ... not really being used in any effective way.

PR #207, opened 2020-11-24
do not let kernel args be captured vars that are lvalue expressions t…

gregrodgers:
…o a pointer.  Change the generated type to getNonLValueExprType and skip the generation of the pointer reference

This is the fix for Devito. Where they declare a pointer with __attribute__ ((alligned(64)))

### DIFF
--- a/clang/lib/CodeGen/CGStmtOpenMP.cpp
+++ b/clang/lib/CodeGen/CGStmtOpenMP.cpp
@@ -481,6 +481,13 @@ static llvm::Function *emitOutlinedFunctionPrologue(
         if (LLVM_AS == LangAS::Default)
           LLVM_AS = LangAS::cuda_device;
         ArgType = Ctx.getAddrSpaceQualType(ArgType, LLVM_AS);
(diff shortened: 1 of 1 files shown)


JonChesterfield:
I think the devito code uses a pointer to an array of arrays, where both array sizes are runtime values. A vla of vla. That's either a compiler extension or a missing error in sema and I'm not clear which. So this patch may be fixing up the result of miscompiling undefined code. It may be better to ask devito to use a type like float [] instead.

ronlieb:
The patch for this in aomp11 breaks 2 tests in examples/openmp

 Return Code for host_function: 2
 Return Code for host_varfn_function: 2

commit 820b593f264abc661c0add184420d2714d2c9063
Author: Greg Rodgers <gregrodgers@users.noreply.github.com>
Date:   Tue Nov 24 13:46:50 2020 -0600

    do not let kernel args be captured vars that are lvalue expressions to a pointer.  Change the generated type to getNonLValueExprType and skip the generation of the pointer reference (#207)

PR #208, opened 2020-11-24
We cannot map VLAs for openmp.

gregrodgers:
This fixes target_teams_reduction so that it fails without compiler crash.  It should fail because we do not support variable length arrays.

### DIFF
--- a/clang/lib/Basic/Targets/AMDGPU.cpp
+++ b/clang/lib/Basic/Targets/AMDGPU.cpp
@@ -312,6 +312,8 @@ AMDGPUTargetInfo::AMDGPUTargetInfo(const llvm::Triple &Triple,
   HasLegalHalfType = true;
   HasFloat16 = true;
 
+  VLASupported = false;
+
   // Set pointer width and alignment for target address space 0.
   PointerWidth = PointerAlign = DataLayout->getPointerSizeInBits();
   if (getMaxPointerWidth() == 64) {


ronlieb (approved):

JonChesterfield:
This disables VLAs everywhere on our target. That seems too aggressive of it's only openmp map clauses that misbehave for them. Can we fatel_error just before we would fail to map instead?

gregrodgers:
This does cause a fatal error instead of a compiler crash.  I changed the test case to check for the fatal error.   Here is the new makefile:

AOMP ?= /usr/lib/aomp

compile.stderr: target_teams_reduction.cpp
        $(AOMP)/bin/aompcc $^ -o $@ 2>compile.stderr >compile.stdout ; true

run: compile.stderr
        grep -q "variable length arrays are not supported" $^
clean:
        @[ -f ./compile.stderr ] && rm ./compile.stderr ; true
        @[ -f ./compile.stdout ] && rm ./compile.stdout ; true

gregrodgers:
I guess we need to see if VLAs work for HIP.

gregrodgers:
This compiles for HIP but aborts in execution.  
```
__global__ void writeIndex(int *b, int n) {
  int i = hipBlockIdx_x;
  int *A_VLA = new int[i]();
  for (int k=0; k<i; k++){
   A_VLA[k] = k;
  }
  if (i < n) {
    b[i] = i;
  }
}
```
VLASupported is false for nvptx64.  So I am comfortable setting it false for amdgcn.   For now it appears that the only place checking this is sema for openmp.   
`

gregrodgers:
OK, after further discussion and enlightenment .  We cannot set VLASupported = false for amdgpu because it will cause some compile fails with hip kernels.  Also, we should be able to make the smoke test target_teams_reduction work.  We need to fix the assert fail during codegen. 

This PR will be deleted.

PR #209, opened 2020-11-28
changes to hostrpc to support fprintf

gregrodgers:


JonChesterfield:
Sure, pretty much a clone of printf. I think we should implement printf on the device side as a call to fprintf(stdout,...), but that can be done sometime later.

PR #213, opened 2020-12-02
Drop parlev field

JonChesterfield:
Missed from previous patch to remove nested parallel


openmp/libomptarget/deviceRTLs/common/omptarget.h
```diff
@@ -323,13 +322,8 @@ extern DEVICE omptarget_nvptx_SimpleMemoryManager
     omptarget_nvptx_simpleMemoryManager;
 extern DEVICE SHARED uint32_t usedMemIdx;
 extern DEVICE SHARED uint32_t usedSlotIdx;
-#ifdef __AMDGCN__
```
JonChesterfield, line 326: drive by fix if (p) a a => a

PR #214, opened 2020-12-02
we are switching the build of openmp runtime and libomptarget to use …

gregrodgers:
…clang. This PR fixes all the warnings that result from using clang

REVIEW QUICKLY.   This is going to be merged for AOMP 11.12-0 release.


openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
```diff
@@ -862,7 +863,7 @@ const Elf64_Sym *elf_lookup(Elf *elf, char *base, Elf64_Shdr *section_hash,
   return nullptr;
 }
 
-typedef struct {
+typedef struct symbol_info {
```
JonChesterfield, line 866: this is a weird thing to warn about

PR #217, opened 2020-12-03
Abstract over device_environment differences between deviceRTLs

JonChesterfield:
With this change, the host runtime patches around differences in the device_environment variable within the deviceRTL in various repos. One of approximately three patches needed to use exactly the same host plugin with trunk, aomp, rocm, hopefully marking an end to implementation divergence.

Copies to the image if the symbol is in data, otherwise copies after loading onto the device.

Accepts that the deviceRTL object may be missing fields. Trunk currently only has debug_level, and will shortly have device_num, but num_devices has unknown eta.

### DIFF
--- a/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
+++ b/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
@@ -845,6 +845,7 @@ const Elf64_Sym *elf_lookup(Elf *elf, char *base, Elf64_Shdr *section_hash,
 typedef struct symbol_info {
   void *addr = nullptr;
   uint32_t size = UINT32_MAX;
+  uint32_t sh_type = SHT_NULL;
 } symbol_info;
 
 int get_symbol_info_without_loading(Elf *elf, char *base, const char *symname,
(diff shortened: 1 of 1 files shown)


openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
```diff
@@ -867,8 +868,23 @@ int get_symbol_info_without_loading(Elf *elf, char *base, const char *symname,
     return 1;
   }
 
-  res->size = static_cast<uint32_t>(sym->st_size);
+  if (sym->st_shndx == SHN_UNDEF) {
```
JonChesterfield, line 871: probably should have treated an undef symbol as an error previously

openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
```diff
@@ -981,42 +1090,18 @@ __tgt_target_table *__tgt_rtl_load_binary_locked(int32_t device_id,
...
-                                             img_size, device_env_Name, &si);
-    if (rc != 0) {
-      DP("Finding global device environment '%s' - symbol missing.\n",
-         device_env_Name);
-      // no need to return FAIL, consider this is a not a device debug build.
-      return ATMI_STATUS_SUCCESS;
-    }
-    if (si.size != sizeof(host_device_env)) {
-      return ATMI_STATUS_ERROR;
+  {
```
JonChesterfield, line 1093: this reduction in scope induces extra `atmi_status_t err =` statements below

JonChesterfield:
This applies poorly to amd-stg-open, as amd-stg-open is missing some changes from trunk (and could do with another clang-format), so won't be proposed to amd-stg-open until it catches up to trunk.

JonChesterfield:
Implemented on trunk/amd-stg-open/aomp-dev

PR #218, opened 2020-12-03
Fix race. Zero init array once, not once per image

JonChesterfield:
Patch fixes exactly this issue and makes no other changes.

The device_State array is allocated once, when the first gpu image is loaded as that is when the required size is known. This allocation is reused for all images on the same gpu, as it can be, and that decreases memory footprint considerably.

Prior to this patch, the array was zero-initialised after copying the pointer to the gpu image. This means it is zero-initialised once per image, so if a previous kernel is still using the storage at the time, it'll be clobbered by loading the next image. This bug is latent.

The fix is to zero initialise once. Load binary is called under a mutex. To keep the patch clean, this is done by writing atmi_calloc in terms of atmi_malloc and hsa_amd_memory_fill. The naming and API is chosen for consistency with the existing memory routines.

### DIFF
--- a/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
+++ b/openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
@@ -946,6 +946,26 @@ __tgt_target_table *__tgt_rtl_load_binary(int32_t device_id,
   return res;
 }
 
+static atmi_status_t atmi_calloc(void **ret_ptr, size_t size,
+                                 atmi_mem_place_t place) {
+  uint64_t rounded = 4 * ((size + 3) / 4);
+  void *ptr;
+  atmi_status_t err = atmi_malloc(&ptr, rounded, place);
+  if (err != ATMI_STATUS_SUCCESS) {
+    return err;
+  }
+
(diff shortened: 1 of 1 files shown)


openmp/libomptarget/plugins/amdgpu/src/rtl.cpp
```diff
@@ -1080,13 +1100,6 @@ __tgt_target_table *__tgt_rtl_load_binary_locked(int32_t device_id,
       fprintf(stderr, "memcpy install of state_ptr failed\n");
       return NULL;
     }
-
-    assert((device_State_bytes & 0x3) == 0); // known >= 4 byte aligned
```
JonChesterfield, line 1084: This wasn't checking alignment, and shouldn't have been checking alignment, merely that the size was a multiple of four. This bug was introduced during a last minute fix to get memory consumption within reasonable bounds for a release.

JonChesterfield:
@estewart08 I'm not sure where you are in the release cycle, please feel free to land this on my behalf if applicable.

estewart08 (approved):

gregrodgers (approved):

PR #221, opened 2020-12-04
Tolerate missing hostrpc library

JonChesterfield:
Provides weak versions of the hostrpc api. If added to trunk, which doesn't have the hostrpc available, these weak versions would be called for init/terminate and assign_buffer is only called if the deviceRTL has tried to call invoke, which is not presently in the trunk deviceRTL.

The extra if null check provides a better user experience than dereferencing null on the gpu, but is only reachable if the deviceRTL assumes a hostrpc is available and the host runtime doesn't have one.


ronlieb (approved):

JonChesterfield:
Gerrit 451584. Only really useful once applied to trunk.

JonChesterfield:
This will be obsolete, change brought in from amd-stg-open

PR #222, opened 2020-12-04
Accept pointer or array for deviceState

JonChesterfield:
nvptx and trunk amdgcn use a global array for deviceState. aomp and rocm use a void *, allocated and initialized by the runtime.

This change restructures the setup of that deviceState by assuming that a symbol greater in size than a void * is already an appropriate global array.

Checked against a deviceRTL that uses the global array - still works, still slower than the alloc+zero strategy.


JonChesterfield:
Gerrit 451651

PR #223, opened 2020-12-10
Add additional path to INSTALL_RPATH for amdpgu plugin.

estewart08:
The debug build of the amdgpu plugin needs to find libhsa-runtime64.so, which resides in ORIGIN/../lib.
This can also be resolved in the future by adding debug builds of rocr/roct to lib-debug or
statically linking them into the plugin.


estewart08:
This fixes the issue with flags1.

JonChesterfield:
Yep, thank you. Sorry for missing the difference in paths for debug vs release.

